PR #6, opened 2023-04-19
More samples

cretz:
## What was changed

* Updated to alpha3
* Added simple activity sample
* Added ASP.NET sample
* Added encryption sample


src/AspNet/Worker/MyWorkflow.workflow.cs
cretz, line 1: Notice my `.workflow.cs` file naming pattern. This is kinda common in .NET to allow different analyzer rules for workflows (as suggested in primary SDK README docs).

src/AspNet/Worker/Program.cs
cretz, line 1: This is the "right" way to run a Temporal worker in .NET probably

src/ActivityWorker/Program.cs
bergundy, line 1: nit: The folder name here doesn't imply that this is a polyglot sample, maybe give that hint somehow in case users look for something like this.

src/Encryption/CodecServer/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        var app = builder.Build();
+
+        // We need CORS so that the browser can access this endpoint from a
+        // different origin
+        app.UseCors(
+            builder => builder.
+                WithHeaders("content-type", "x-namespace").
+                WithMethods("POST").
+                WithOrigins("http://localhost:8080", "https://cloud.temporal.io/"));
```
bergundy, line 27: Might want to put `8233` here for the CLI's web UI and document that users may need to customize these URLs.

tests/Program.cs
```diff
@@ -0,0 +1,23 @@
...
+public static class Program
+{
+    internal static bool InProc { get; private set; }
+
+    internal static bool Verbose { get; private set; }
+
+    public static int Main(string[] args)
+    {
+        InProc = true;
+        Verbose = args.Contains("-verbose");
```
bergundy, line 12: Is it standard to use single dash flags in .NET?

bergundy: Seems inconsistent with `--help` below.

bergundy (approved):

src/Encryption/CodecServer/Program.cs
cretz: :+1: Will do

tests/Program.cs
cretz: This isn't my flag FWIW, it's just me wrapping XUnit's manual runner. I only wrapped it because the standard `dotnet test` runs a separate process that does not show live output well from CLI among other issues. Most won't use this.

src/Encryption/CodecServer/Program.cs
cretz: Fixed

PR #16, opened 2023-05-30
Activity dependency injection sample

### LINKED ISSUE #9: Sample request: Activity DI

People need to access their DI-based objects inside an activity and it may not be obvious how to do this manually.

May also be related to https://github.com/temporalio/sdk-dotnet/issues/46

cretz:
## What was changed

An activity dependency injection sample was added that shows how to use the `ActivityDefinition.Create`'s custom `invoker` with a service provider to support activity injection.

This is fairly simple/small code, but the question is whether any of this needs to be incorporated into the primary SDK. If we want to avoid DI dependency in primary SDK we have these options:

1. Leave as simple sample people can take code from
2. Support a `TemporalWorkerOptions.AddAllActivities<T>(IServiceProvider)` method or some other way to provide late-bound instantiator
3. Create a new project for `Temporalio.Extensions.DependencyInjection` with basically the code seen here

## Checklist

1. Closes #9
5. Relates to https://github.com/temporalio/sdk-dotnet/issues/66


prom3theu5:
Option 3 is the best way to ship this imho
Net Core is designed by nature to be an open extensible framework, and I think having the Extensions lib would be the way to go - it matches what people are probably used to by now.

It also opens up a sub "family" of lib support, i.e.

 Temporalio.Extensions.DependencyInjection
 Temporalio.Extensions.DependencyInjection.Autofac
 Temporalio.Extensions.DependencyInjection.SimpleInjector
 Temporalio.Extensions.DependencyInjection.Lamar
 
 Etc
 
 Thanks for this - being completely new to Temporal and thinking about saga's as workflows, i'd been lagging a bit in terms of clearly thinking how to implement DI

tgrieger-sf:
Genuine question, is there a reason to choose this path over having dependency injection "just work"? For example, when injecting a dependency into a grain in Orleans, nothing special has to be done with the Grain in order to make that happen. Could the same be done with Activities?

cretz:
> Genuine question, is there a reason to choose this path over having dependency injection "just work"? For example, when injecting a dependency into a grain in Orleans, nothing special has to be done with the Grain in order to make that happen. Could the same be done with Activities?

Yes, we plan on making it just work. For constructor-level injection, we want users to be able to control the instantiation of activity classes instead of requiring dependency injection. But we will make it just work with the `Temporalio.Extensions.Hosting` project so it will work just like https://orleanscn.github.io/orleans/Documentation/Advanced-Concepts/Dependency-Injection.html. For method-level injection, we want the method signature to be the same for implementers as it is for callers. This is important for type safety for the caller side (we don't want to end up in a situation like [durable entities](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-dotnet-entities#accessing-entities-through-interfaces) where an interface is required).

tgrieger-sf:
> Yes, we plan on making it just work. For constructor-level injection, we want users to be able to control the instantiation of activity classes instead of requiring dependency injection. But we will make it just work with the `Temporalio.Extensions.Hosting` project so it will work just like https://orleanscn.github.io/orleans/Documentation/Advanced-Concepts/Dependency-Injection.html. For method-level injection, we want the method signature to be the same for implementers as it is for callers. This is important for type safety for the caller side (we don't want to end up in a situation like [durable entities](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-dotnet-entities#accessing-entities-through-interfaces) where an interface is required).
> 
> I probably won't even merge this sample because it'll be moot after that project appears soon.

That makes total sense to me, thanks for the reply.

cretz:
Closing. This no longer applies now that we have a `Temporalio.Extensions.Hosting` package

Issue #18, opened 2023-06-05
[Bug] Remove "Ref" comment


cretz:
### Describe the bug

At least at https://github.com/temporalio/samples-dotnet/blob/580cac9b8165fb49f98d20acd70a96bf27e04a0c/src/ActivitySimple/MyWorkflow.workflow.cs#L13, there is still a comment to `Ref`.

PR #23, opened 2023-06-16
Remove comment about Ref

jakejscott:
## What was changed
Removed Ref comments

1. Closes https://github.com/temporalio/samples-dotnet/issues/18

2. How was this tested:
No need

3. Any docs updates needed?
N/A


cretz (approved):
Thanks!

Issue #11, opened 2023-04-25
Sample request: Activity heartbeating and cancellation


cretz:
### Describe the solution you'd like

Need to show heartbeating and cancellation handling in an activity

PR #25, opened 2023-06-17
Activities heart beating and cancellation

jakejscott:
## What was changed
Add activities heart beating and cancellation sample

## Why?
To show how to use heart beating and cancellation

## Checklist

1. Closes https://github.com/temporalio/samples-dotnet/issues/11

2. How was this tested:
Running it locally

3. Any docs updates needed?
Updated readme

### DIFF
--- a/src/ActivityHeartbeatingCancellation/MyActivities.cs
+++ b/src/ActivityHeartbeatingCancellation/MyActivities.cs
@@ -1,0 +1,37 @@
+using Microsoft.Extensions.Logging;
+using Temporalio.Activities;
+using Temporalio.Exceptions;
+
+namespace TemporalioSamples.ActivityHeartbeatingCancellation;
+
+public static class MyActivities
+{
+    [Activity]
+    public static async Task FakeProgressAsync(int sleepIntervalMs = 1000)
+    {
+        try
+        {
+            // Allow for resuming from heartbeat
+            var startingPoint = ActivityExecutionContext.Current.Info.HeartbeatDetails.Any()
+                ? await ActivityExecutionContext.Current.Info.HeartbeatDetailAtAsync<int>(0)
+                : 1;
+
+            ActivityExecutionContext.Current.Logger.LogInformation("Starting activity at progress: {StartingPoint}", startingPoint);
+
+            for (var progress = startingPoint; progress <= 100; ++progress)
+            {
+                ActivityExecutionContext.Current.CancellationToken.ThrowIfCancellationRequested();
+                ActivityExecutionContext.Current.Logger.LogInformation("Progress: {Progress}", progress);
+                await Task.Delay(sleepIntervalMs);
+                ActivityExecutionContext.Current.Heartbeat(progress);
+            }
+
+            ActivityExecutionContext.Current.Logger.LogInformation("Fake progress activity completed");
+        }
+        catch (OperationCanceledException e) when (e.InnerException is CancelledFailureException)
+        {
+            ActivityExecutionContext.Current.Logger.LogInformation("Fake progress activity cancelled");
+            throw;
+        }
+    }
+}
--- a/src/ActivityHeartbeatingCancellation/MyActivities.cs
+++ b/src/ActivityHeartbeatingCancellation/MyActivities.cs
@@ -1,6 +1,5 @@
 using Microsoft.Extensions.Logging;
 using Temporalio.Activities;
-using Temporalio.Exceptions;
 
 namespace TemporalioSamples.ActivityHeartbeatingCancellation;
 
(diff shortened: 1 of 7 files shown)


src/ActivityHeartbeatingCancellation/MyActivities.cs
```diff
@@ -0,0 +1,37 @@
...
+            {
+                ActivityExecutionContext.Current.CancellationToken.ThrowIfCancellationRequested();
+                ActivityExecutionContext.Current.Logger.LogInformation("Progress: {Progress}", progress);
+                await Task.Delay(sleepIntervalMs);
+                ActivityExecutionContext.Current.Heartbeat(progress);
+            }
+
+            ActivityExecutionContext.Current.Logger.LogInformation("Fake progress activity completed");
+        }
+        catch (OperationCanceledException e) when (e.InnerException is CancelledFailureException)
```
cretz, line 31:
```suggestion
        catch (OperationCanceledException)
```

I think demonstrating just catching the cancel is good enough and probably better practice than relying on the cause

src/ActivityHeartbeatingCancellation/MyActivities.cs
```diff
@@ -0,0 +1,37 @@
...
+                ? await ActivityExecutionContext.Current.Info.HeartbeatDetailAtAsync<int>(0)
+                : 1;
+
+            ActivityExecutionContext.Current.Logger.LogInformation("Starting activity at progress: {StartingPoint}", startingPoint);
+
+            for (var progress = startingPoint; progress <= 100; ++progress)
+            {
+                ActivityExecutionContext.Current.CancellationToken.ThrowIfCancellationRequested();
+                ActivityExecutionContext.Current.Logger.LogInformation("Progress: {Progress}", progress);
+                await Task.Delay(sleepIntervalMs);
```
cretz, line 25:
```suggestion
                await Task.Delay(sleepIntervalMs, ActivityExecutionContext.Current.CancellationToken);
                ActivityExecutionContext.Current.Logger.LogInformation("Progress: {Progress}", progress);
```

I think this better demonstrates the use of the cancellation token

src/ActivityHeartbeatingCancellation/MyWorkflow.workflow.cs
```diff
@@ -0,0 +1,30 @@
...
+                new()
+                {
+                    StartToCloseTimeout = TimeSpan.FromMinutes(5),
+                    HeartbeatTimeout = TimeSpan.FromSeconds(3),
+                    // Don't send rejection to our Workflow until the Activity has confirmed cancellation
+                    CancellationType = ActivityCancellationType.WaitCancellationCompleted,
+                });
+        }
+        catch (ActivityFailureException e) when (e.InnerException is CancelledFailureException)
+        {
```
cretz, line 26:
```suggestion
        catch (TemporalException e) when (TemporalException.IsCancelledException(e))
        {
            // This catches the cancel just for demonstration, you usually don't want to catch it
```

I think this is a better sample showing how to catch cancel generally and also adds a comment making it clear that usually you don't want to catch it

src/ActivityHeartbeatingCancellation/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+    await handle.CancelAsync();
+    Console.WriteLine("Cancelled workflow successfully");
+
+    try
+    {
+        await handle.GetResultAsync();
+    }
+    catch (WorkflowFailedException e) when (e.InnerException is CancelledFailureException)
+    {
+        Console.WriteLine("await handle.GetResultAsync() threw because Workflow was cancelled");
```
cretz, line 64: You swallow the cancel in the workflow, so will this ever occur? Maybe rethrow from the workflow's catch clause?

src/ActivityHeartbeatingCancellation/MyActivities.cs
```diff
@@ -0,0 +1,36 @@
...
+            // Allow for resuming from heartbeat
+            var startingPoint = ActivityExecutionContext.Current.Info.HeartbeatDetails.Any()
+                ? await ActivityExecutionContext.Current.Info.HeartbeatDetailAtAsync<int>(0)
+                : 1;
+
+            ActivityExecutionContext.Current.Logger.LogInformation("Starting activity at progress: {StartingPoint}", startingPoint);
+
+            for (var progress = startingPoint; progress <= 100; ++progress)
+            {
+                ActivityExecutionContext.Current.CancellationToken.ThrowIfCancellationRequested();
```
cretz, line 22: I don't think you need this line

src/ActivityHeartbeatingCancellation/MyActivities.cs
```diff
@@ -0,0 +1,36 @@
...
+                ActivityExecutionContext.Current.Logger.LogInformation("Progress: {Progress}", progress);
```
cretz, line 23: I think this log line should go after the sleep where you are recording the progress

cretz (commented):
Just a couple more minor things and we're good to merge. Thanks!

cretz (approved):
Thanks!

PR #31, opened 2023-07-19
Worker Versioning and Update SDK

Sushisource:
Update SDK and add worker versioning


cretz (approved):
@Sushisource - Updated SDK. Can you confirm everything still looks good from your end? If so, feel free to merge.

PR #32, opened 2023-09-02
Workflow Mutex Sample

devbased:
## What was changed
Added a new sample showcasing how to implement a Mutex Workflow for managing concurrent access to a shared resource.

~_NOTE: Since the required SDK version (the one with [workflow history info](https://github.com/temporalio/sdk-dotnet/commit/9b927098be7e277d688798b0da235fd2182643ca)) hasn't been published yet, I had to add links to the SDK in TemporalioSamples.sln and Directory.build.props_~

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -19,9 +19,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Temporalio" Version="0.1.0-beta1" />
-    <PackageReference Include="Temporalio.Extensions.Hosting" Version="0.1.0-beta1" />
-    <PackageReference Include="Temporalio.Extensions.OpenTelemetry" Version="0.1.0-beta1" />
+    <PackageReference Include="Temporalio" Version="0.1.0-beta2" />
+    <PackageReference Include="Temporalio.Extensions.Hosting" Version="0.1.0-beta2" />
+    <PackageReference Include="Temporalio.Extensions.OpenTelemetry" Version="0.1.0-beta2" />
     <!--
     Can also reference the SDK downloaded to a local directory:
     <ProjectReference Include="$(MSBuildThisFileDirectory)..\temporal-sdk-dotnet\src\Temporalio\Temporalio.csproj" />
--- a/src/Mutex/Activities.cs
+++ b/src/Mutex/Activities.cs
@@ -1,0 +1,38 @@
+namespace TemporalioSamples.Mutex;
+
+using Temporalio.Activities;
+
+public record NotifyLockedInput(string ResourceId, string ReleaseSignalName);
+
+public record UseApiThatCantBeCalledInParallelInput(TimeSpan SleepFor);
+
+public record NotifyUnlockedInput(string ResourceId);
+
+public static class Activities
+{
+    [Activity]
+    public static void NotifyLocked(NotifyLockedInput input)
+    {
+        ActivityExecutionContext.Current.Logger.LogInformation(
+            "Lock for resource '{ResourceId}' acquired, release signal name '{ReleaseSignalName}'", input.ResourceId, input.ReleaseSignalName);
+    }
+
+    [Activity]
+    public static async Task UseApiThatCantBeCalledInParallelAsync(UseApiThatCantBeCalledInParallelInput input)
+    {
+        var logger = ActivityExecutionContext.Current.Logger;
+
+        logger.LogInformation("Sleeping for '{SleepFor}'...", input.SleepFor);
+
+        await Task.Delay(input.SleepFor);
+
+        logger.LogInformation("Done sleeping!");
+    }
+
+    [Activity]
+    public static void NotifyUnlocked(NotifyUnlockedInput input)
+    {
+        ActivityExecutionContext.Current.Logger.LogInformation(
+            "Lock for resource '{ResourceId}' released", input.ResourceId);
+    }
+}
--- a/src/Mutex/Impl/ILockHandle.cs
+++ b/src/Mutex/Impl/ILockHandle.cs
@@ -1,0 +1,10 @@
+namespace TemporalioSamples.Mutex.Impl;
+
+public interface ILockHandle : IAsyncDisposable
+{
+    public string LockInitiatorId { get; }
+
+    public string ResourceId { get; }
+
+    public string ReleaseSignalName { get; }
+}
--- a/src/Mutex/Impl/LockRequest.cs
+++ b/src/Mutex/Impl/LockRequest.cs
@@ -1,0 +1,3 @@
+namespace TemporalioSamples.Mutex.Impl;
+
+internal record LockRequest(string InitiatorId, string AcquireLockSignalName, TimeSpan? Timeout = null);
--- a/src/Mutex/Program.cs
+++ b/src/Mutex/Program.cs
@@ -1,0 +1,75 @@
+﻿using System.Diagnostics;
+using Temporalio.Client;
+using Temporalio.Worker;
+using TemporalioSamples.Mutex;
+using TemporalioSamples.Mutex.Impl;
+
+var client = await TemporalClient.ConnectAsync(new("localhost:7233")
+{
+    LoggerFactory = LoggerFactory.Create(builder =>
+        builder.AddSimpleConsole(options => options.TimestampFormat = "[HH:mm:ss] ").SetMinimumLevel(LogLevel.Information)),
+});
+
+async Task RunWorkerAsync()
+{
+    // Cancellation token cancelled on ctrl+c
+    using var tokenSource = new CancellationTokenSource();
+    Console.CancelKeyPress += (_, eventArgs) =>
+    {
+        tokenSource.Cancel();
+        eventArgs.Cancel = true;
+    };
+
+    // Run worker until cancelled
+    Console.WriteLine("Running worker");
+
+    using var worker = new TemporalWorker(
+        client,
+        new TemporalWorkerOptions(taskQueue: "workflow-mutex-sample")
+            .AddWorkflowMutex(client)
+            .AddAllActivities(typeof(Activities), null)
+            .AddWorkflow<WorkflowWithMutex>());
+    try
+    {
+        await worker.ExecuteAsync(tokenSource.Token);
+    }
+    catch (OperationCanceledException)
+    {
+        Console.WriteLine("Worker cancelled");
+    }
+}
+
+async Task ExecuteWorkflowsWithMutex(string resourceId)
+{
+    await Task.WhenAll(Execute(), Execute());
+
+    return;
+
+    async Task Execute()
+    {
+        var workflowId = "test-" + Guid.NewGuid();
+        Console.WriteLine($"Starting test workflow with id '{workflowId}'. Connecting to lock workflow '{resourceId}'");
+
+        var sw = Stopwatch.StartNew();
+        var handle = await client.StartWorkflowAsync(
+            (WorkflowWithMutex wf) => wf.RunAsync(new WorkflowWithMutexInput(resourceId, TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(7.5))),
+            new(workflowId, "workflow-mutex-sample"));
+
+        Console.WriteLine($"Test workflow '{workflowId}' started");
+
+        await handle.GetResultAsync();
+        Console.WriteLine($"Test workflow '{workflowId}' finished after {sw.ElapsedMilliseconds}ms");
+    }
+}
+
+switch (args.ElementAtOrDefault(0))
+{
+    case "worker":
+        await RunWorkerAsync();
+        break;
+    case "workflow":
+        await ExecuteWorkflowsWithMutex(args.ElementAtOrDefault(1) ?? "locked-resource-id");
+        break;
+    default:
+        throw new ArgumentException("Must pass 'worker' or 'workflow' as the first argument");
+}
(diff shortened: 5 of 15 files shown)


Directory.Build.props
```diff
@@ -19,13 +19,15 @@
...
-    Can also reference the SDK downloaded to a local directory:
+    <!--<PackageReference Include="Temporalio" Version="0.1.0-beta1" />-->
+    <!--<PackageReference Include="Temporalio.Extensions.Hosting" Version="0.1.0-beta1" />-->
+    <!--<PackageReference Include="Temporalio.Extensions.OpenTelemetry" Version="0.1.0-beta1" />-->
+    
+    <!--Can also reference the SDK downloaded to a local directory:-->
     <ProjectReference Include="$(MSBuildThisFileDirectory)..\temporal-sdk-dotnet\src\Temporalio\Temporalio.csproj" />
-    -->
+    <ProjectReference Include="$(MSBuildThisFileDirectory)..\temporal-sdk-dotnet\src\Temporalio.Extensions.Hosting\Temporalio.Extensions.Hosting.csproj" />
+    <ProjectReference Include="$(MSBuildThisFileDirectory)..\temporal-sdk-dotnet\src\Temporalio.Extensions.OpenTelemetry\Temporalio.Extensions.OpenTelemetry.csproj" />
```
cretz, line 29: :+1: Will leave this PR unmerged until the next SDK release

src/Mutex/Impl/ILockHandle.cs
```diff
@@ -0,0 +1,10 @@
+namespace TemporalioSamples.Mutex.Impl;
```
cretz, line 1: I think `ILockHandle` and `WorkflowMutex` can move up to the `TemporalioSamples.Mutex` with the rest (can make former inner class of latter if wanting fewer files). And can probably move `MutexWorkflow.workflow.cs` up there too and put most things it uses as inner classes if we want to avoid file sprawl.

Your call, but I think there's value in fewer files/namespaces here

src/Mutex/Impl/Internal/MutexWorkflow.workflow.cs
```diff
@@ -0,0 +1,55 @@
+using Temporalio.Workflows;
+
+namespace TemporalioSamples.Mutex.Impl.Internal;
+
+[Workflow]
+internal class MutexWorkflow
+{
+    private readonly ILockHandler lockHandler = WorkflowMutex.CreateLockHandler();
```
cretz, line 8: Not sure there is value for a separate lock handler concept instead of just embedding logic in this class which is the only place that uses it

src/Mutex/Impl/Internal/MutexActivities.cs
```diff
@@ -0,0 +1,35 @@
...
+    public async Task SignalWithStartMutexWorkflowAsync(SignalWithStartMutexWorkflowInput input)
+    {
+        var activityInfo = ActivityExecutionContext.Current.Info;
+
+        await this.client.StartWorkflowAsync(
+            (MutexWorkflow mw) => mw.RunAsync(null),
+            new WorkflowOptions(input.ResourceId, activityInfo.TaskQueue)
+            {
+                StartSignal = RequestLockSignalName,
+                StartSignalArgs = new object[] { new LockRequest(activityInfo.WorkflowId, input.AcquireLockSignalName, input.LockTimeout), },
```
cretz, line 32: I'm thinking I should make a more typesafe form of this as a helper on the options

src/Mutex/Impl/Internal/MutexActivities.cs
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    [Activity]
+    public async Task SignalWithStartMutexWorkflowAsync(SignalWithStartMutexWorkflowInput input)
+    {
+        var activityInfo = ActivityExecutionContext.Current.Info;
+
+        await this.client.StartWorkflowAsync(
+            (MutexWorkflow mw) => mw.RunAsync(null),
+            new WorkflowOptions(input.ResourceId, activityInfo.TaskQueue)
```
cretz, line 29: I think the workflow ID should have something prepended just in case they are using this resource ID as a workflow ID elsewhere

src/Mutex/Impl/Internal/MutexWorkflow.workflow.cs
```diff
@@ -0,0 +1,55 @@
...
+    private readonly Queue<LockRequest> requests = new();
+
+    [WorkflowRun]
+    public async Task RunAsync(MutexWorkflowInput? input = null)
```
cretz, line 12: Don't have to do it, but there's a neat feature in .NET SDK not in other SDKs - you can have a constructor with `[WorkflowInit]` that accepts this input there

cretz, line 12: Often clearer to accept the non-null input and make your caller just pass an empty form of it. You're the only caller of this, so you can pass an empty input.

src/Mutex/Impl/Internal/MutexWorkflowInput.cs
```diff
@@ -0,0 +1,3 @@
+namespace TemporalioSamples.Mutex.Impl.Internal;
+
+internal record MutexWorkflowInput(IEnumerable<LockRequest> InitialRequests);
```
cretz, line 3:
```suggestion
internal record MutexWorkflowInput(IReadOnlyCollection<LockRequest> InitialRequests);
```

I have been discouraging enumerable as parameters in this use case because of the confusion with users when they expect it to be lazy or not re-enumerable.

src/Mutex/Impl/Internal/MutexWorkflow.workflow.cs
```diff
@@ -0,0 +1,55 @@
...
+            if (requests.Count == 0)
+            {
+                logger.LogInformation("No lock requests, waiting for more...");
+
+                await Workflow.WaitConditionAsync(() => requests.Count > 0);
+            }
+
+            while (requests.TryDequeue(out var lockRequest))
+            {
+                await this.lockHandler.HandleAsync(lockRequest);
```
cretz, line 32: So it looks like this can only handle one lock at a time which I am not sure is what we want

src/Mutex/README.md
cretz, line 1: Don't forget to update main README with reference to this one

src/Mutex/WorkflowWithMutex.workflow.cs
```diff
@@ -0,0 +1,41 @@
+﻿using Temporalio.Workflows;
+using TemporalioSamples.Mutex.Impl;
+
+namespace TemporalioSamples.Mutex;
+
+[Workflow]
+public class WorkflowWithMutex
+{
+    private readonly ActivityOptions activityOptions = new()
```
cretz, line 9: Might as well be static

src/Mutex/WorkflowWithMutexInput.cs
```diff
@@ -0,0 +1,3 @@
+namespace TemporalioSamples.Mutex;
+
+public record WorkflowWithMutexInput(string ResourceId, int SleepForMs = 500, int LockTimeoutMs = 1000);
```
cretz, line 3: Feel free to accept `TimeSpan`s here if you are willing to pass them around for other use cases

cretz (commented):
Added some comments, but overall this is the same as I was thinking. I like the idea of wrapping as a utility.

src/Mutex/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    var mutexActivities = new MutexActivities(client);
+
+    using var worker = new TemporalWorker(
+        client,
+        new TemporalWorkerOptions(taskQueue: "workflow-mutex-sample")
+            .AddAllActivities(typeof(Activities), null)
+            .AddAllActivities(mutexActivities)
+            .AddWorkflow<WorkflowWithMutex>()
+            .AddWorkflow<MutexWorkflow>());
```
cretz, line 34: Feel free to have a helper or extension method for registering mutex stuffs instead of a user having to know the things needed

src/Mutex/Impl/Internal/MutexWorkflow.workflow.cs
devbased: Here a separate object is used, which is returned from WorkflowMutex for the reason that signals need to be sent and they are not strongly typed. Therefore, I thought that it would be better for all internal communication to be centralized in WorkflowMutex. In general, when I was writing this sample, I constantly felt a contradiction, that in some places we have strongly typed calls, while others do not have such capability

src/Mutex/Impl/ILockHandle.cs
devbased: I wanted everything related to the implementation of the mutex to be clearly separated, meaning that potentially it could be a separate project, where only WorkflowMutex.LockAsync (and possibly methods that register workflows/activities) is accessible. On the other hand, since this is just an example, creating multiple projects is overkill

src/Mutex/Impl/Internal/MutexWorkflow.workflow.cs
devbased: I understood that the constructor is called with an input, and you can populate the Queue there, but RunAsync still needs to have an argument, which is not used if there is a constructor. It looks strange in the end

src/Mutex/Impl/ILockHandle.cs
cretz: Hrmm, yeah, makes sense to separate everything out, just maybe not two levels deep then. I wish I could think of a better word than `Impl` heh.

src/Mutex/Impl/Internal/MutexWorkflow.workflow.cs
cretz: Hrmm, yeah it makes sense to me since you're sharing those names. I think I was just hoping to have to jump around to fewer files for this simple sample to see what things do.

src/Mutex/Impl/Internal/MutexWorkflow.workflow.cs
cretz: Concur, but it should be noted that "signal with start" in Temporal, for special ordering reasons, will call a signal handler _before_ "run" is called. But in your case you only ever send signal-with-start with no initial requests, so you're good I think.

src/Mutex/Impl/Internal/MutexWorkflow.workflow.cs
cretz: Ah, true, I was thinking of a case where a mutex workflow could lock several independent resources, which doesn't apply here. Can consider this resolved. I think I was confused by the presence of the `ResourceId` in the lock request, which seems like it should be a product of the workflow as a whole.

src/Mutex/NotifyLockedInput.cs
cretz, line 1: To reduce file count, can put these tiny classes as inner classes of the activity. Not sure I ever would in real code, just trying to trim down the many files people have to jump across to read this sample

cretz (approved):
Looks good. Marking approved. I think it can be simplified a good bit, though I know that file and package sprawl is kinda the .NET way. Approving for CI run, but of course it will fail due to the relative-path dep, no worry there, we can leave this PR open for a bit.

devbased:
@cretz I reduced the number of files, but I didn't refactor the inputs into nested classes; instead, I placed them outside in the same file where they are being used. Additionally, I provided a bit more detailed description of the project's structure in the README.md

src/Mutex/Activities.cs
```diff
@@ -0,0 +1,38 @@
+using Temporalio.Activities;
+
+namespace TemporalioSamples.Mutex;
```
cretz, line 3: Ideally the namespace is at the top, but don't think it matters that much.

src/Mutex/Impl/Internal/MutexActivities.cs
cretz: I opened https://github.com/temporalio/sdk-dotnet/issues/134

src/Mutex/Impl/LockRequest.cs
```diff
@@ -0,0 +1,3 @@
+namespace TemporalioSamples.Mutex.Impl;
+
+internal record LockRequest(string InitiatorId, string ResourceId, string AcquireLockSignalName, TimeSpan? Timeout = null);
```
cretz, line 3: Can we remove the `ResourceId` from this since it's not a property of the request it's a property of the entire workflow? If the resource ID is needed from the workflow, can store on the workflow or derive from workflow ID.

cretz (approved):
Looks good!

> I didn't refactor the inputs into nested classes; instead, I placed them outside in the same file where they are being used.

I think it's clearer as nested classes (I like the type-per-file rule), but not a big deal I suppose.

Will merge when SDK is updated. Thanks!

src/Mutex/Activities.cs
devbased: Currently, `.editorconfig` does not contain `csharp_using_directive_placement`. Should I add `csharp_using_directive_placement = inside_namespace`?

devbased: Cleaned up 'usings'

cretz:
Thanks for the update! Gonna merge #39 first if that's ok (unsure if it'll merge cleanly).

cretz:
Resolved conflict (using GH's built-in conflict editor for the first time, was interesting). Merging once CI passes. Thanks again!

PR #46, opened 2023-12-11
feat: add saga sample

### LINKED ISSUE #33: Sample request: Saga

Basically a simple saga sample following principles in other SDKs and outlined at https://temporal.io/blog/compensating-actions-part-of-a-complete-breakfast-with-sagas and https://temporal.io/blog/saga-pattern-made-easy

aboelkassem:
## What was changed
Added a saga pattern sample in dotnet

## Why?
To handle compensation errors in microservices using temporal.io

## Checklist

1. Closes #33


src/Saga/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+    try
+    {
+        await worker.ExecuteAsync(tokenSource.Token);
+    }
+    catch (OperationCanceledException)
+    {
+        Console.WriteLine("Worker cancelled");
+    }
+}
+async Task ExecuteWorkflowAsync()
```
cretz, line 40:
```suggestion

async Task ExecuteWorkflowAsync()
```

src/Saga/Saga.cs
```diff
@@ -0,0 +1,56 @@
+﻿using Microsoft.Extensions.Logging;
+
+namespace TemporalioSamples.Saga;
+
+public class Saga
```
cretz, line 5: I think this may be adding more features than is needed. I think this should follow the TypeScript sample, see https://github.com/temporalio/samples-typescript/tree/main/saga. Basically just put a `List<Func<Task>> compensations = new()` at the top of the function needing compensations (i.e. the workflow function), then `foreach (var comp in compensations.Reverse() {` in the `catch`. I don't think on-compensation-complete or on-compensation-error callbacks are needed in the sample either, users can implement those as they need in their code. Users may have different things they need to do (or not do) in each situation.

I think there's value in showing that "saga" is not a special thing in Temporal, anyone can run any code in catch. There's no need to break this out.

aboelkassem: @cretz fixed

src/Saga/SagaWorkflow.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            StartToCloseTimeout = TimeSpan.FromSeconds(90), // schedule a retry if the Activity function doesn't return within 90 seconds
+            RetryPolicy = new()
+            {
+                InitialInterval = TimeSpan.FromSeconds(15), // first try will occur after 15 seconds
+                BackoffCoefficient = 1, // double the delay after each retry
+                MaximumInterval = TimeSpan.FromMinutes(1), // up to a maximum delay of 1 minute
+                MaximumAttempts = 2, // fail the Activitiesivity after 2 attempts
+            },
+        };
```
cretz, line 25: We usually don't want to encourage people to set all of these. The defaults are ideal, especially for samples. If there are specific ones for compensation, so be it, but I'd rather this be:

```suggestion
        var options = new ActivityOptions() { StartToCloseTimeout = TimeSpan.FromSeconds(90) };
```

src/Saga/SagaWorkflow.cs
```diff
@@ -0,0 +1,55 @@
...
+            // throw new Exception
+            await Workflow.ExecuteActivityAsync(() => Activities.StepWithError(transfer), options);
+        }
+        catch (Exception)
+        {
+            logger.LogInformation("Exception caught. Initiating compensation...");
+            compensations.Reverse();
+            foreach (var comp in compensations)
+            {
+                await comp.Invoke();
```
cretz, line 50: You can silently overwrite exceptions here. I think you should match the TypeScript sample and explicitly log-and-swallow any compensation failures

src/Saga/README.md
cretz, line 1: Can you also update the primary README.md to mention this sample similar to the other ones

aboelkassem: @cretz thanks for your valuable comments 
they have been noted

src/Saga/SagaWorkflow.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                try
+                {
+                    await comp.Invoke();
+                }
+                catch (Exception ex)
+                {
+                    Workflow.Logger.LogError("failed to compensate: {Message}", ex.Message);
+                    // swallow errors
+                    throw;
```
cretz, line 57:
```suggestion
```

You wouldn't rethrow if you want to swallow

src/Saga/SagaWorkflow.cs
```diff
@@ -0,0 +1,62 @@
...
+            compensations.Reverse();
+            foreach (var comp in compensations)
+            {
+                try
+                {
+                    await comp.Invoke();
+                }
+                catch (Exception ex)
+                {
+                    Workflow.Logger.LogError("failed to compensate: {Message}", ex.Message);
```
cretz, line 55:
```suggestion
                    Workflow.Logger.LogError(ex, "Failed to compensate");
```

Capitalize log message and properly pass exception to logger

cretz (commented):
Looking good, just a couple of minor things. Feel free to add a test if you want, but not required.

README.md
```diff
@@ -24,6 +24,7 @@ Prerequisites:
 * [Schedules](src/Schedules) - How to schedule workflows to be run at specific times in the future.
 * [WorkerSpecificTaskQueues](src/WorkerSpecificTaskQueues) - Use a unique task queue per Worker to have certain Activities only run on that specific Worker.
 * [WorkerVersioning](src/WorkerVersioning) - How to use the Worker Versioning feature to more easily deploy changes to Workflow & other code.
+* [Saga](src/Saga) - Demonstrates how to implement a saga pattern.
```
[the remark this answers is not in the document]
aboelkassem: done

aboelkassem:
@cretz The problem here with writing tests is since I deleted the retry policy and will take the defaults, so `StepWithError` will be running forever until succeeds.

What do you suggest?

cretz:
> @cretz The problem here with writing tests is since I deleted the retry policy it will take the default values, so StepWithError will be running forever until succeeds.

Change `throw new ApplicationFailureException("Simulated failure");` to `throw new ApplicationFailureException("Simulated failure", nonRetryable: true);`

src/Saga/SagaWorkflow.cs
```diff
@@ -0,0 +1,61 @@
...
+        if (compensations.Count > 0)
+        {
+            compensations.Reverse();
+            foreach (var comp in compensations)
+            {
+                try
+                {
+                    await comp.Invoke();
+                }
+                catch (ApplicationException ex)
```
cretz, line 53: Why'd this change from `Exception`? This will never be reached (may be worth tests to confirm) since activity failures are always wrapped in activity exceptions.

aboelkassem: @cretz 
if i removed `throw;` while catching generic Exception, the compiler/cod analyser raise the following error 
`CA1031	Modify 'CompensateAsync' to catch a more specific allowed exception type, or rethrow the exception`

cretz: You can `#pragma warning disable CA1031` + `#pragma warning restore CA1031` with a comment explaining why. Here's an example: https://github.com/temporalio/sdk-dotnet/blob/b64dd6def30060ca29aabd3bfec70848ac36d052/src/Temporalio/Converters/DefaultFailureConverter.cs#L84-L102. Sometimes it's ok to override these analyzer checks if we know what we're doing.

aboelkassem: Fixed

Thank you @cretz for all your valuable reviews, really appreciate it

aboelkassem: for sure, I will write tests for it later when I have extra time

cretz: Happy to help! Thanks for sticking with it.

cretz (approved):
Approved pending CI

src/Saga/SagaWorkflow.cs
```diff
@@ -0,0 +1,63 @@
...
+            logger.LogInformation("Exception caught. Initiating compensation...");
+            await CompensateAsync(compensations);
+            throw;
+        }
+    }
+
+    private async Task CompensateAsync(List<Func<Task>> compensations)
+    {
+        if (compensations.Count > 0)
+        {
```
cretz, line 45:
```suggestion
```

Don't really need to check that it's non-empty

Issue #47, opened 2023-12-13
Fix aspnet DI sample


jakejscott:
### Is your feature request related to a problem? Please describe.

Update sample to use DI without awaiting the clientTask.

PR #49, opened 2023-12-13
Fix asp.net sample to use DI

jakejscott:
Fixes DI for the asp.net sample

1. Closes https://github.com/temporalio/samples-dotnet/issues/47

2. How was this tested:
Manually by running the web and worker and hitting http://localhost:5000

3. Any docs updates needed?
N/A


cretz:
This was already using DI. Based on https://stackoverflow.com/questions/43240405/async-provider-in-net-core-di and others I've seen, they suggest that async be deferred until needed hence why I put the task in the container (though in retrospect maybe it should have been a lazy task).

I wonder if  `TemporalClientAdapter` is something we should add to `Temporalio.Extensions.Hosting`. But it'd just be a lazy wrapper.

Issue #50, opened 2024-01-23
[Bug] AesGcm instance should not be reused in EncryptionCodec


cretz:
### Describe the bug

AesGcm is not thread safe, so it is dangerous to share in the codec. See https://github.com/dotnet/runtime/discussions/71198 and other places.

PR #52, opened 2024-02-01
Makes AesGcm Sample Thread Safe

joelrozen:
## What was changed

Changed Encryption Codec in Encryption Sample to use a new instance of AesGcm on each call.

## Why?

See #50 

## Checklist

1. Closes  #50 

2. How was this tested:
 Ran several thousand simultaneous workflows

3. Any docs updates needed? No

### DIFF
--- a/src/Encryption/Codec/EncryptionCodec.cs
+++ b/src/Encryption/Codec/EncryptionCodec.cs
@@ -16,7 +16,6 @@ public sealed class EncryptionCodec : IPayloadCodec, IDisposable
 
     private readonly byte[] KeyBytes;
     private readonly ByteString keyIDByteString;
-    private readonly AesGcm aes;
 
     public EncryptionCodec(string keyID = DefaultKeyID, string? key = DefaultKey)
     {
--- a/src/Encryption/Codec/EncryptionCodec.cs
+++ b/src/Encryption/Codec/EncryptionCodec.cs
@@ -17,11 +17,11 @@ public sealed class EncryptionCodec : IPayloadCodec
     private readonly byte[] key;
     private readonly ByteString keyIDByteString;
 
-    public EncryptionCodec(string keyID = DefaultKeyID, byte[]? _key = DefaultKey)
+    public EncryptionCodec(string keyID = DefaultKeyID, byte[]? key)
     {
         KeyID = keyID;
         keyIDByteString = ByteString.CopyFromUtf8(keyID);
-        key = _key;
+        this.key = key ?? DefaultKey;
     }
 
     public string KeyID { get; private init; }
--- a/src/Encryption/Codec/EncryptionCodec.cs
+++ b/src/Encryption/Codec/EncryptionCodec.cs
@@ -8,20 +8,21 @@ using Temporalio.Converters;
 public sealed class EncryptionCodec : IPayloadCodec, IDisposable
 {
     public const string DefaultKeyID = "test-key-id";
-    public static readonly byte[] DefaultKey = System.Text.Encoding.ASCII.GetBytes("test-key-test-key-test-key-test!");
+    private static string DefaultKey = "test-key-test-key-test-key-test!";
     // Taken from other language samples for compatibility
     private const int NonceSize = 12;
     private const int TagSize = 16;
     private static readonly ByteString EncodingByteString = ByteString.CopyFromUtf8("binary/encrypted");
 
+    private readonly byte[] KeyBytes;
     private readonly ByteString keyIDByteString;
     private readonly AesGcm aes;
 
-    public EncryptionCodec(string keyID = DefaultKeyID, byte[]? key = null)
+    public EncryptionCodec(string keyID = DefaultKeyID, string? key = DefaultKey)
     {
         KeyID = keyID;
         keyIDByteString = ByteString.CopyFromUtf8(keyID);
-        aes = new(key ?? DefaultKey);
+        KeyBytes = System.Text.Encoding.ASCII.GetBytes(key);
     }
 
     ~EncryptionCodec() => Dispose();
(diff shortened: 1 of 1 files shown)


src/Encryption/Codec/EncryptionCodec.cs
```diff
@@ -8,20 +8,20 @@ namespace TemporalioSamples.Encryption.Codec;
 public sealed class EncryptionCodec : IPayloadCodec, IDisposable
 {
     public const string DefaultKeyID = "test-key-id";
-    public static readonly byte[] DefaultKey = System.Text.Encoding.ASCII.GetBytes("test-key-test-key-test-key-test!");
+    private static string DefaultKey = "test-key-test-key-test-key-test!";
```
cretz, line 11: Should be no need to change this I figure, let's keep keys as byte arrays from a user POV (they don't have to have string representations)

src/Encryption/Codec/EncryptionCodec.cs
```diff
@@ -30,7 +30,6 @@ public EncryptionCodec(string keyID = DefaultKeyID, byte[]? key = null)
 
     public void Dispose()
```
cretz, line 31: Can just remove this call, destructor, and `IDisposable` from this class I believe.

src/Encryption/Codec/EncryptionCodec.cs
```diff
@@ -8,20 +8,20 @@ namespace TemporalioSamples.Encryption.Codec;
...
 {
     public const string DefaultKeyID = "test-key-id";
-    public static readonly byte[] DefaultKey = System.Text.Encoding.ASCII.GetBytes("test-key-test-key-test-key-test!");
+    private static string DefaultKey = "test-key-test-key-test-key-test!";
     // Taken from other language samples for compatibility
     private const int NonceSize = 12;
     private const int TagSize = 16;
     private static readonly ByteString EncodingByteString = ByteString.CopyFromUtf8("binary/encrypted");
 
+    private readonly byte[] KeyBytes;
```
cretz, line 17:
```suggestion
    private readonly byte[] key;
```

No need to capitalize the private field and can just consider keys to always be byte arrays and not have/require a string form.

cretz (commented):
Thanks! A couple of suggestions but otherwise looks good.

src/Encryption/Codec/EncryptionCodec.cs
```diff
@@ -14,26 +14,18 @@ public sealed class EncryptionCodec : IPayloadCodec, IDisposable
     private const int TagSize = 16;
     private static readonly ByteString EncodingByteString = ByteString.CopyFromUtf8("binary/encrypted");
 
+    private readonly byte[] key;
     private readonly ByteString keyIDByteString;
-    private readonly AesGcm aes;
 
-    public EncryptionCodec(string keyID = DefaultKeyID, byte[]? key = null)
+    public EncryptionCodec(string keyID = DefaultKeyID, byte[]? _key = DefaultKey)
```
cretz, line 20: You can't set a default parameter to a non-constant value I don't believe (this is breaking CI). I think you should leave this as it is, and change below to `this.key = key ?? DefaultKey`.

cretz (commented):
Looks good! Just a compile error to fix.

src/Encryption/Codec/EncryptionCodec.cs
```diff
@@ -14,26 +14,18 @@ public sealed class EncryptionCodec : IPayloadCodec, IDisposable
     private const int TagSize = 16;
     private static readonly ByteString EncodingByteString = ByteString.CopyFromUtf8("binary/encrypted");
 
+    private readonly byte[] key;
     private readonly ByteString keyIDByteString;
-    private readonly AesGcm aes;
 
-    public EncryptionCodec(string keyID = DefaultKeyID, byte[]? key = null)
+    public EncryptionCodec(string keyID = DefaultKeyID, byte[]? key)
```
cretz, line 20:
```suggestion
    public EncryptionCodec(string keyID = DefaultKeyID, byte[]? key = null)
```

This is still breaking compilation because the default value was removed.

src/Encryption/Codec/EncryptionCodec.cs
```diff
@@ -79,15 +71,20 @@ private byte[] Encrypt(byte[] data)
...
     }
 
     private byte[] Decrypt(byte[] data)
     {
         var bytes = new byte[data.Length - NonceSize - TagSize];
-        aes.Decrypt(
-            data.AsSpan(0, NonceSize), data.AsSpan(NonceSize + TagSize), data.AsSpan(NonceSize, TagSize), bytes.AsSpan());
-        return bytes;
+
+        using (var aes = new AecGcm(key)) {
```
cretz, line 84:
```suggestion
        using (var aes = new AesGcm(key))
        {
```

src/Encryption/Codec/EncryptionCodec.cs
```diff
@@ -79,15 +71,20 @@ private byte[] Encrypt(byte[] data)
         RandomNumberGenerator.Fill(nonceSpan);
 
         // Perform encryption
-        aes.Encrypt(nonceSpan, data, bytes.AsSpan(NonceSize + TagSize), bytes.AsSpan(NonceSize, TagSize));
-        return bytes;
+        using (var aes = new AesGcm(key)) {
```
cretz, line 74:
```suggestion
        using (var aes = new AesGcm(key))
        {
```

cretz (commented):
CI still not passing, a couple of typos. Run `dotnet build` to confirm it builds.

joelrozen:
image

cretz (approved):
Thanks!

PR #57, opened 2024-03-28
Add timer sample

lorensr:
Thinking maybe we should add handling workflow cancellation to this sample. Wdyt, and would this be the right way?

```
namespace TemporalioSamples.Timer;

using Microsoft.Extensions.Logging;
using Temporalio.Workflows;
using System.Threading.Tasks;

[Workflow]
public class MyWorkflow
{
    [WorkflowRun]
    public async Task RunAsync()
    {
        try
        {
            while (true)
            {
                Workflow.CancellationToken.ThrowIfCancellationRequested();

                await Workflow.DelayAsync(1000, Workflow.CancellationToken);

                var result = await Workflow.ExecuteActivityAsync(
                    () => MyActivities.Charge(user),
                    new() { StartToCloseTimeout = TimeSpan.FromMinutes(5) },
                    Workflow.CancellationToken); // Pass the cancellation token

                Workflow.Logger.LogInformation("Activity result: {Result}", result);
            }
        }
        catch (OperationCanceledException)
        {
            // Handle any cleanup upon cancellation here
            Workflow.Logger.LogInformation("Workflow cancelled.");
        }
    }
}
```


src/Timer/Subscription.workflow.cs
```diff
@@ -0,0 +1,22 @@
...
+
+[Workflow]
+public class Subscription
+{
+    [WorkflowRun]
+    public async Task RunAsync(string userId)
+    {
+        while (true)
+        {
+            await Workflow.DelayAsync(TimeSpan.FromDays(30));
```
cretz, line 14: Arguably it's not worth a whole example to demonstrate using `DelayAsync` (I am not sure any other SDK has such a simple timer sample)

src/Timer/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+// Create a client to localhost on default namespace
+var client = await TemporalClient.ConnectAsync(new("localhost:7233")
+{
+    LoggerFactory = LoggerFactory.Create(builder =>
+        builder.
+            AddSimpleConsole(options => options.TimestampFormat = "[HH:mm:ss] ").
+            SetMinimumLevel(LogLevel.Information)),
+});
+
+async Task RunWorkerAsync()
```
cretz, line 15: We haven't on most samples, but we may find using https://github.com/temporalio/sdk-dotnet/tree/main/src/Temporalio.Extensions.Hosting for worker running to be easier in our samples here and elsewhere (this is what many of our users use)

cretz (commented):
> Thinking maybe we should add handling workflow cancellation to this sample. Wdyt, and would this be the right way?

The workflow cancellation token is already the default cancellation token for these calls if none provided so you don't have to change any workflow code.

src/Timer/Subscription.workflow.cs
lorensr: I think it will be helpful to some, particularly those new to temporal. I'd point someone to this to show what I meant by "sleep for days" while pitching.

I also added a bit on handling workflow cancellation

src/Timer/Program.cs
lorensr: Making worker file shorter sgtm, going to leave this for future work for time's sake.

src/Timer/Subscription.workflow.cs
cretz: There is a case where too many samples of too trivial things can be present. Not sure other SDK samples repositories would have accepted such trivial timer sample, unsure if .NET should be the inconsistent leader here.

lorensr: All but .net/python have at least one mention of timer in the readme's list of samples, and I didn't find a use of delayasync in this repo. I think basic samples are valuable, and imagine they're easy to skip over for those who are more experienced and know what a timer is.

src/Timer/Subscription.workflow.cs
```diff
@@ -0,0 +1,32 @@
...
+            {
+                await Workflow.DelayAsync(TimeSpan.FromDays(30));
+
+                var result = await Workflow.ExecuteActivityAsync(
+                    () => MyActivities.Charge(userId),
+                    new() { StartToCloseTimeout = TimeSpan.FromMinutes(5) });
+                Workflow.Logger.LogInformation("Activity result: {Result}", result);
+            }
+        }
+        catch (OperationCanceledException)
```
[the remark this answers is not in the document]
lorensr: Thanks!

src/Timer/Subscription.workflow.cs
cretz: :+1: Those other samples do seem more involved (maybe w/ the exception of Go). Would you be willing to make this sample at least as robust as the TypeScript one?

lorensr: Would like to, but don't have time to, sorry!

cretz (approved):
OK, can merge. I have opened https://github.com/temporalio/samples-dotnet/issues/59 for finishing out the sample.

PR #58, opened 2024-03-28
Add SignalsQueries sample

lorensr:
Couple TODOs left, one with question

### DIFF
--- a/src/SignalsQueries/LoyaltyProgram.workflow.cs
+++ b/src/SignalsQueries/LoyaltyProgram.workflow.cs
@@ -3,10 +3,12 @@ namespace TemporalioSamples.SignalsQueries;
 using Microsoft.Extensions.Logging;
 using Temporalio.Workflows;
 
+public record Purchase(string Id, int TotalCents);
+
 [Workflow]
 public class LoyaltyProgram
 {
-    private string? userId;
+    private readonly Queue<Purchase> toProcess = new();
 
     [WorkflowQuery]
     public int Points { get; private set; }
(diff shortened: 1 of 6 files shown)


src/SignalsQueries/LoyaltyProgram.workflow.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    private string? userId;
+    private int points = 0;
+
+    [WorkflowRun]
+    public async Task RunAsync(string userId)
+    {
+        this.userId = userId;
+
+        // TODO dear chad, how do i await cancellation? or something short/simple that prevents from returning
```
cretz, line 17: If you want to wait forever in and SDK, you can do a wait condition on false. In .NET this is `await Workflow.WaitConditionAsync(() => false);`

src/SignalsQueries/LoyaltyProgram.workflow.cs
```diff
@@ -0,0 +1,38 @@
...
+            Workflow.Logger.LogInformation("Sending coupon to {UserId}", userId);
+            await Workflow.ExecuteActivityAsync(
+                () => MyActivities.SendCoupon(userId),
+                new() { ScheduleToCloseTimeout = TimeSpan.FromMinutes(5) });
+            points -= 10_000;
+        }
+    }
+
+    [WorkflowQuery]
+    public int GetPoints() => points;
```
cretz, line 37: Wanna see something cool in .NET SDK? We support queries as properties (but not signals), so you can remove the `private int points = 0;` and change this to:

```csharp
[WorkflowQuery]
public int Points { get; private set; }
```

src/SignalsQueries/Program.cs
```diff
@@ -0,0 +1,65 @@
...
+    // Cancellation token cancelled on ctrl+c
+    using var tokenSource = new CancellationTokenSource();
+    Console.CancelKeyPress += (_, eventArgs) =>
+    {
+        tokenSource.Cancel();
+        eventArgs.Cancel = true;
+    };
+
+    // Create an activity instance with some state
+    var activities = new MyActivities();
```
cretz, line 26: Unused variable

src/SignalsQueries/Program.cs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+async Task ExecuteWorkflowAsync()
+{
+    Console.WriteLine("Executing workflow");
+    await client.ExecuteWorkflowAsync(
+        (LoyaltyProgram wf) => wf.RunAsync("user-id-123"),
+        new(id: "signals-queries-workflow-id", taskQueue: "signals-queries-sample"));
+
+    // TODO signal and query
```
cretz, line 52: Will wait on this to approve

cretz (commented):
Looking good, welcome to .NET 😁

cretz (approved):
Would usually want tests for this, but this is good even without them (we don't have a strong precedent on requiring tests for samples). Still, may want a test to confirm behavior at some point.

PR #73, opened 2024-06-24
Safe message handler sample

cretz:
## What was changed

Sample showing atomic message handling in .NET w/ `Semaphore` using the logic from https://github.com/temporalio/samples-python/pull/123


tests/WorkflowEnvironment.cs
cretz, line 1: Added this abstraction as part of this PR so we can reuse dev servers across tests

tests/WorkflowEnvironment.cs
```diff
@@ -0,0 +1,36 @@
...
+    public async Task InitializeAsync()
+    {
+        env = await Temporalio.Testing.WorkflowEnvironment.StartLocalAsync(new()
+        {
+            DevServerOptions = new()
+            {
+                ExtraArgs =
+                [
+                    "--dynamic-config-value",
+                    "frontend.enableUpdateWorkflowExecution=true",
```
Quinn-With-Two-Ns, line 23: I wonder if we should enable stuff like this when go to public preview?

cretz: Yes, I think we should according to the implied opt-in vs opt-out nature of release stages at https://docs.temporal.io/evaluate/release-stages. But of course that is a server thing and not related to this PR.

cretz:
By request, I am not merging this until https://github.com/temporalio/samples-python/pull/123 is merged.

src/SafeMessageHandlers/README.md
```diff
@@ -0,0 +1,17 @@
+# Safe Message Handlers
+
+This sample shows a workflow using `Temporalio.Workflows.Semaphore` to atomically process certain blocks of workflow
+code to prevent data race issues. The sample code demonstrates assigning cluster nodes to jobs atomically.
```
Sushisource, line 4: For some strict definition of data race, that can't happen in Temporal regardless (what's being prevented here is I think more strictly a logical race). I'd suggest we say "to serialize execution of certain blocks of workflow code" just to be extra clear here.

src/SafeMessageHandlers/ClusterManagerWorkflow.workflow.cs
```diff
@@ -0,0 +1,209 @@
...
+                Select(kvp => kvp.Key).
+                ToList();
+            if (unassignedNodes.Count < input.NumNodes)
+            {
+                throw new ApplicationFailureException(
+                    $"Cannot allocate {input.NumNodes} nodes, have only {unassignedNodes.Count} available");
+            }
+            var assignedNodes = unassignedNodes[..input.NumNodes];
+            // This await would be dangerous without nodesLock because it yields control and allows
+            // interleaving
```
Sushisource, line 119:
```suggestion
            // This await would be dangerous without nodesLock because it yields control and would
            // allow execution to interleave with other concurrent updates
```

Sushisource (approved):
Cool sample. Just some language suggestions.

cretz:
Merging now without all handler stuff to not wait on next release. We can come back and add later if we want.

PR #77, opened 2024-07-23
Added Counter Interceptor Sample

rross:
## What was changed
Added a CounterInterceptor example

## Why?
Demonstrate how to use interceptors with the .NET SDK

## Checklist

1. Closes n/a

2. How was this tested:
Added a test that validates the counts.

3. Any docs updates needed?
Updated the main README to include this sample


README.md
```diff
@@ -17,6 +17,7 @@ Prerequisites:
 * [ActivityWorker](src/ActivityWorker) - Use .NET activities from a workflow in another language.
 * [AspNet](src/AspNet) - Demonstration of a generic host worker and an ASP.NET workflow starter.
 * [ClientMtls](src/ClientMtls) - How to use client certificate authentication, e.g. for Temporal Cloud.
+* [CounterInterceptor](src/CounterInterceptor/) - Simple Workflow and Client Interceptors example.
```
cretz, line 20: Alphabetical (needs to be one lower)

src/CounterInterceptor/ClientCounter.cs
```diff
@@ -0,0 +1,89 @@
...
+
+using System.Numerics;
+
+public class ClientCounter
+{
+    private const string NumberOfWorkflowExecutions = "numOfWorkflowExec";
+    private const string NumberOfSignals = "numOfSignals";
+    private const string NumberOfQueries = "numOfQueries";
+    private static Dictionary<string, Dictionary<string, BigInteger?>> perWorkflowIdDictionary =
+        new Dictionary<string, Dictionary<string, BigInteger?>>();
```
cretz, line 11:
```suggestion
        new();
```

src/CounterInterceptor/ClientCounter.cs
```diff
@@ -0,0 +1,89 @@
+namespace TemporalioSamples.CounterInterceptor;
+
+using System.Numerics;
+
+public class ClientCounter
+{
+    private const string NumberOfWorkflowExecutions = "numOfWorkflowExec";
+    private const string NumberOfSignals = "numOfSignals";
+    private const string NumberOfQueries = "numOfQueries";
+    private static Dictionary<string, Dictionary<string, BigInteger?>> perWorkflowIdDictionary =
```
cretz, line 10:
```suggestion
    private static Dictionary<string, Dictionary<string, uint64>> perWorkflowIdDictionary =
```

I think `uint64` is enough. If you incremented the counter every nanosecond (not very possible), it'd take you over 580 years to overflow.

cretz, line 10:
```suggestion
    private static Dictionary<string, WorkflowCounts> workflowCounts = new();
```

And have a mutable record like:

```public record WorkflowCounts
{
    uint64 Executions { get; internal set; }
    uint64 Signals { get; internal set; }
    uint64 Queries { get; internal set; }
}
```

Could add a `ToString()` on it if you must have multiline strings though may not be needed. And just call the field `workflowCounts`. You don't have to name the field by what the key is and what the value is. I haven't tested any of this, but can be cleaned up a bit

src/CounterInterceptor/ClientCounter.cs
```diff
@@ -0,0 +1,89 @@
...
+        new Dictionary<string, Dictionary<string, BigInteger?>>();
+
+    public static string Info()
+    {
+        string result = string.Empty;
+        foreach (var item in perWorkflowIdDictionary)
```
cretz, line 16: Use linq. Assuming you have the previously mentioned record, this is probably good enough: `string.Join("\n", perWorkflowIdDictionary.Select(kvp -> "** Workflow ID: {kvp.Key}, Counts: {kvp.Value}"))` or something like that

src/CounterInterceptor/WorkerCounter.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+        else
+        {
+            var current = value[type];
+            var next = current + 1;
+            value[type] = next;
+        }
+    }
+
+    public static BigInteger? NumOfWorkflowExecutions(string workflowId)
```
cretz, line 36: Unused helpers

src/CounterInterceptor/ClientCounter.cs
```diff
@@ -0,0 +1,89 @@
...
+                "\n** Workflow ID: " + item.Key +
+                "\n\tTotal Number of Workflow Exec: " + info[NumberOfWorkflowExecutions] +
+                "\n\tTotal Number of Signals: " + info[NumberOfSignals] +
+                "\n\tTotal Number of Queries: " + info[NumberOfQueries];
+        }
+
+        return result;
+    }
+
+    public static BigInteger? NumOfWorkflowExecutions(string workflowId)
```
cretz, line 29: Assuming the field is properly changed to `private static Dictionary<string, WorkflowCounts> workflowCounts`, a simple `public static WorkflowCounts? GetWorkflowCounts(string workflowId) => workflowCounts.TryGetValue(workflowId, out var counts) ? counts : null` will work

src/CounterInterceptor/ClientCounter.cs
```diff
@@ -0,0 +1,89 @@
...
+        if (value[type] == null)
+        {
+            value[type] = 1;
+        }
+        else
+        {
+            var current = value[type];
+            var next = current + 1;
+            value[type] = next;
+        }
```
fairlydurable, line 87: Not super familiar with the language, but couldn't this be expressed with a ternary? 

```
value[type] = (value[type] == null) ? 1 : value[type] + 1;
```

src/CounterInterceptor/Constants.cs
cretz, line 1: This file is likely not needed. Non-test code only uses each of these constants in one file, so put them there (test code should be using more random names to avoid clashing with other test runs)

src/CounterInterceptor/MyChildWorkflow.workflow.cs
```diff
@@ -0,0 +1,21 @@
...
+    };
+
+    [WorkflowRun]
+    public async Task<string> ExecChildAsync(string name, string title)
+    {
+        string result = await Workflow.ExecuteActivityAsync((MyActivities act) => act.SayHello(name, title), activityOptions);
+        result += await Workflow.ExecuteActivityAsync((MyActivities act) => act.SayGoodBye(name, title), activityOptions);
+
+        return result;
+    }
```
cretz, line 20:
```suggestion
    public Task<string> ExecChildAsync(string name, string title) =>
        await Workflow.ExecuteActivityAsync(
            (MyActivities act) => act.SayHello(name, title), activityOptions) +
        await Workflow.ExecuteActivityAsync(
            (MyActivities act) => act.SayGoodBye(name, title), activityOptions);
```

This reads clearer to me

src/CounterInterceptor/MyWorkflow.workflow.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    private string name = string.Empty;
+    private string title = string.Empty;
+    private bool exit; // automatically defaults to false
+
+    [WorkflowRun]
+    public async Task<string> ExecAsync()
+    {
+        // wait for greeting info
+        await Workflow.WaitConditionAsync(() => name != null && title != null);
```
cretz, line 16: This condition always passes

src/CounterInterceptor/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+            return value;
+        }
+
+        var address = GetEnvVarWithDefault("TEMPORAL_ADDRESS", "127.0.0.1:7233");
+        var temporalNamespace = GetEnvVarWithDefault("TEMPORAL_NAMESPACE", "default");
+        var tlsCertPath = GetEnvVarWithDefault("TEMPORAL_TLS_CERT", string.Empty);
+        var tlsKeyPath = GetEnvVarWithDefault("TEMPORAL_TLS_KEY", string.Empty);
+        TlsOptions? tls = null;
+        if (!string.IsNullOrEmpty(tlsCertPath) && !string.IsNullOrEmpty(tlsKeyPath))
```
fairlydurable, line 25: I'm a little concerned about edge cases. If tlsCert/Key are not set, should the logic continue or should it be an early fail path. Since this is "main" and (I assume) running as an app, why not leave an error message and exit with an abnormal status.

I'd rather see `tls` set in a happy path after early conditions/assertions are all met for successfully moving forward.

Apologies if I'm off-base.

src/CounterInterceptor/MyWorkflow.workflow.cs
```diff
@@ -0,0 +1,56 @@
...
+    public string QueryName()
+    {
+        return name;
+    }
+
+    [WorkflowQuery]
+    public string QueryTitle()
+    {
+        return title;
+    }
```
cretz, line 49: Remove the private fields and use properties. Queries are built to be property getters:

```suggestion
    [WorkflowQuery]
    public string Name { get; private set; } = string.Empty;

    [WorkflowQuery]
    public string Title { get; private set; } = string.Empty;
```

src/CounterInterceptor/MyWorkflow.workflow.cs
```diff
@@ -0,0 +1,56 @@
...
+    private bool exit; // automatically defaults to false
+
+    [WorkflowRun]
+    public async Task<string> ExecAsync()
+    {
+        // wait for greeting info
+        await Workflow.WaitConditionAsync(() => name != null && title != null);
+
+        // Execute Child Workflow
+        string result = await Workflow.ExecuteChildWorkflowAsync(
```
cretz, line 19:
```suggestion
        var result = await Workflow.ExecuteChildWorkflowAsync(
```

src/CounterInterceptor/MyWorkflow.workflow.cs
```diff
@@ -0,0 +1,56 @@
...
+            (MyChildWorkflow wf) => wf.ExecChildAsync(name, title),
+            new()
+            {
+                Id = Constants.ChildWorkflowId,
+            });
```
cretz, line 24:
```suggestion
            new() { Id = Constants.ChildWorkflowId });
```

Don't have to span lines

src/CounterInterceptor/MyChildWorkflow.workflow.cs
```diff
@@ -0,0 +1,21 @@
...
+[Workflow]
+public class MyChildWorkflow
+{
+    private readonly ActivityOptions activityOptions = new()
+    {
+        StartToCloseTimeout = TimeSpan.FromSeconds(10),
+    };
+
+    [WorkflowRun]
+    public async Task<string> ExecChildAsync(string name, string title)
```
cretz, line 14:
```suggestion
    public async Task<string> RunAsync(string name, string title)
```

Not required obviously, but I think calling all run calls `RunAsync` is clearest

src/CounterInterceptor/MyWorkflow.workflow.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        // Wait for exit signal
+        await Workflow.WaitConditionAsync(() => exit != false);
```
cretz, line 27:
```suggestion
        await Workflow.WaitConditionAsync(() => exit);
```

src/CounterInterceptor/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+            {
+                return defaultValue;
+            }
+            return value;
+        }
+
+        var address = GetEnvVarWithDefault("TEMPORAL_ADDRESS", "127.0.0.1:7233");
+        var temporalNamespace = GetEnvVarWithDefault("TEMPORAL_NAMESPACE", "default");
+        var tlsCertPath = GetEnvVarWithDefault("TEMPORAL_TLS_CERT", string.Empty);
+        var tlsKeyPath = GetEnvVarWithDefault("TEMPORAL_TLS_KEY", string.Empty);
```
cretz, line 23: This is not something we do with our other samples. If/when we come back and make samples work across many environments, we will update them all. No need to have just this sample accept environment variables unlike others.

src/CounterInterceptor/README.md
fairlydurable, line 1: * Vale linter is unhappy about the use of simple that might "come across as condescending".
  I don't personally mind the word but pointing out that you use it twice close together.

* Vale hates 'as well as'. Suggestion: 'and'.

* Vale doesn't like uppercase "Sample". We're weird about upper/lower case. I'd check with another docs person to be sure though.

src/CounterInterceptor/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+            workerOptions);
+
+        // Run worker until cancelled
+        Console.WriteLine("Running worker...");
+        try
+        {
+            // Start the workers
+            var workerResult = worker.ExecuteAsync(tokenSource.Token);
+
+            // start the workflow
```
cretz, line 72: Inconsistent with whether you capitalize comments

src/CounterInterceptor/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            var name = await handle.QueryAsync(wf => wf.QueryName());
+            var title = await handle.QueryAsync(wf => wf.QueryTitle());
+
+            // send exit signal to workflow
+            await handle.SignalAsync(wf => wf.ExitAsync());
+
+            var result = await handle.GetResultAsync();
+
+            Console.WriteLine($"Workflow result is {result}", result);
```
cretz, line 88: Inconsistent with whether you use string concatenation or string interpolation

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
```diff
@@ -0,0 +1,65 @@
...
+        public override Task<WorkflowUpdateHandle<TResult>> StartWorkflowUpdateAsync<TResult>(
+            StartWorkflowUpdateInput input)
+        {
+            // not tracking this
+            return base.StartWorkflowUpdateAsync<TResult>(input);
+        }
+
+        private static string CheckId(string? id)
+        {
+            var returnValue = "None";
```
fairlydurable, line 56: Suggestion: convert entire to a ternary.

src/CounterInterceptor/WorkerCounter.cs
```diff
@@ -0,0 +1,91 @@
+namespace TemporalioSamples.CounterInterceptor;
+
+using System.Numerics;
+
+public static class WorkerCounter
+{
+    public const string NumberOfWorkflowExecutions = "numOfWorkflowExec";
```
fairlydurable, line 7: While I personally prefer camel for readability (and, for constants like this upper camel -- but again, that's personal), in the learn samples we consistently use screaming snake for shared constants.

new TemporalWorkerOptions(taskQueue: "MONEY_TRANSFER_TASK_QUEUE")

Feel free to check with other docs folk

src/CounterInterceptor/WorkerCounter.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    public static void Add(string workflowId, string type)
+    {
+        if (!perWorkflowIdDictionary.TryGetValue(workflowId, out Dictionary<string, BigInteger?>? value))
+        {
+            value = DefaultInfoMap();
+            perWorkflowIdDictionary.Add(workflowId, value);
+        }
+
+        if (value[type] == null)
```
fairlydurable, line 24: See previous comment, for setting value[type] to 1 or current + 1 with ternary.

src/CounterInterceptor/WorkerCounter.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        return perWorkflowIdDictionary[workflowId][NumberOfQueries];
+    }
+
+    public static string Info()
+    {
+        string result = string.Empty;
+        foreach (var item in perWorkflowIdDictionary)
+        {
+            var info = item.Value;
```
fairlydurable, line 66: Suggestion: `itemInfo` vs `info`, tying it to `item` in iteration and showcasing use

tests/CounterInterceptor/MyWorkflowTests.cs
```diff
@@ -0,0 +1,76 @@
...
+            var name = await handle.QueryAsync(wf => wf.QueryName());
+            var title = await handle.QueryAsync(wf => wf.QueryTitle());
+
+            // send exit signal to workflow
+            await handle.SignalAsync(wf => wf.ExitAsync());
+
+            // Wait for the workflow to complete
+            var result = await handle.GetResultAsync();
+
+            // validate that the worker counters have the correct numbers for the parent
```
fairlydurable, line 56: nit: I have an allergy to comments that start with lower case. Casing and grammar reflect professionalism and attention to detail, and enhance readability, which are critical in a professional environment.

README.md
rross: Fixed.

src/CounterInterceptor/ClientCounter.cs
rross: Changed to uint. Had style issues with uint64

src/CounterInterceptor/ClientCounter.cs
rross: Fixed.

src/CounterInterceptor/ClientCounter.cs
rross: Refactoring this to a unit removed the need for the null check

src/CounterInterceptor/Constants.cs
rross: Good call. Removed

src/CounterInterceptor/MyChildWorkflow.workflow.cs
rross: Yeah I agree. Fixed.

src/CounterInterceptor/MyChildWorkflow.workflow.cs
rross: Agree. Fixed.

src/CounterInterceptor/MyWorkflow.workflow.cs
rross: I'm not sure I understand what you mean. Can you clarify please?

src/CounterInterceptor/MyWorkflow.workflow.cs
rross: Fixed.

src/CounterInterceptor/MyWorkflow.workflow.cs
rross: Fixed.

src/CounterInterceptor/MyWorkflow.workflow.cs
rross: Very cool. Fixed.

src/CounterInterceptor/MyWorkflow.workflow.cs
```diff
@@ -0,0 +1,56 @@
...
+    public string QueryTitle()
+    {
+        return title;
+    }
+
+    [WorkflowSignal]
+    public async Task ExitAsync()
+    {
+        this.exit = true;
+    }
```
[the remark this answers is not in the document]
rross: Fixed.

src/CounterInterceptor/Program.cs
rross: I get what you are saying and really wish we did this everywhere...

src/CounterInterceptor/Program.cs
rross: Removed based on Chad's previous comment.

src/CounterInterceptor/MyWorkflow.workflow.cs
cretz: Name and title are never null, so what is the purpose of this condition?

src/CounterInterceptor/Program.cs
cretz: We plan to

src/CounterInterceptor/README.md
rross: Can you provide more context please? Did not find "as well as".
Sample is the standard used for the .NET namespace for these

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
rross: I personally don't like ternary expressions and opt for clarity.

src/CounterInterceptor/WorkerCounter.cs
rross: Removed the constants based on Chad's recommendation

src/CounterInterceptor/WorkerCounter.cs
rross: null check is no longer needed after refactoring to use unit

src/CounterInterceptor/WorkerCounter.cs
rross: Fixed.

tests/CounterInterceptor/MyWorkflowTests.cs
rross: Fixed

src/CounterInterceptor/README.md
fairlydurable: > The sample demonstrates: the use of a simple Worker Workflow Interceptor that counts the number of Workflow Executions, Child Workflow Executions, and Activity Executions **as well as** the number of Signals and Queries. It is based off of the [Java Sample](https://github.com/temporalio/samples-java/tree/main) located

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
fairlydurable: Looks like dotnet has null coalescing of some sort, so that will be even better for making the code more readable and clear. See Chad's comment above.

src/CounterInterceptor/README.md
rross: Thanks. I missed that this was in the README..

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
rross: Fixed.

src/CounterInterceptor/ClientCounter.cs
rross: Fixed. Added a bit of complexity in the Add() method but overall I like it.

src/CounterInterceptor/ClientCounter.cs
rross: Fixed

src/CounterInterceptor/README.md
fairlydurable: Vale is our docs standards linter for prose

src/CounterInterceptor/SimpleCounterWorkerInterceptor.cs
```diff
@@ -0,0 +1,99 @@
...
+            StartChildWorkflowInput input)
+        {
+            WorkerCounter.Add(Workflow.Info.WorkflowId, WorkerCounter.NumberOfChildWorkflowExecutions);
+            return base.StartChildWorkflowAsync<TWorkflow, TResult>(input);
+        }
+    }
+
+    private sealed class ActivityInbound : ActivityInboundInterceptor
+    {
+        private readonly SimpleCounterWorkerInterceptor root;
```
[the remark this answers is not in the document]
rross: Fixed

src/CounterInterceptor/TemporalioSamples.CounterInterceptor.csproj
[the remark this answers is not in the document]
rross: Fixed.

tests/TemporalioSamples.Tests.csproj
```diff
@@ -21,6 +21,7 @@
   <ItemGroup>
     <ProjectReference Include="..\src\ActivitySimple\TemporalioSamples.ActivitySimple.csproj" />
     <ProjectReference Include="..\src\ActivityWorker\TemporalioSamples.ActivityWorker.csproj" />
+    <ProjectReference Include="..\src\CounterInterceptor\TemporalioSamples.CounterInterceptor.csproj" />
```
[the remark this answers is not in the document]
rross: Fixed

tests/CounterInterceptor/MyWorkflowTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        var clientOptions = (TemporalClientOptions)env.Client.Options.Clone();
+        clientOptions.Interceptors = new[]
+        {
+            new SimpleClientCallsInterceptor(),
+        };
+
+        var client = new TemporalClient(env.Client.Connection, clientOptions);
+
+        var workerOptions = new TemporalWorkerOptions(
+            TemporalioSamples.CounterInterceptor.Constants.TaskQueue).
```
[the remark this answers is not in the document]
rross: Fixed

src/CounterInterceptor/ClientCounter.cs
```diff
@@ -0,0 +1,87 @@
+namespace TemporalioSamples.CounterInterceptor;
```
cretz, line 1: A bit surprised this passes CI, maybe we don't have a `dotnet format` check. Usually we want there to be a blank line between namespace declaration and type declarataion

src/CounterInterceptor/ClientCounter.cs
```diff
@@ -0,0 +1,87 @@
...
+public record ClientCounts
```
cretz, line 2: Should not have a top-level type declared in any non-`Program.cs` file that doesn't match its name. Either make it a nested class or give it its own file.

src/CounterInterceptor/ClientCounter.cs
```diff
@@ -0,0 +1,87 @@
...
+
+    public uint Queries { get; internal set; }
+
+    public override string ToString()
+    {
+        return
+                "\n\tTotal Number of Workflow Exec: " + Executions +
+                "\n\tTotal Number of Signals: " + Signals +
+                "\n\tTotal Number of Queries: " + Queries;
+    }
```
cretz, line 16:
```suggestion
    public override string ToString() =>
        $"\n\tTotal Number of Workflow Exec: {Executions}\n\t" +
        $"Total Number of Signals: {Signals}\n\tTotal Number of Queries: {Queries}";
    }
```

Can break up long line, but might as well use interpolation, and might as well use `=>` for single-line methods. Granted, despite what I said before, using `ToString()` for multiline strings only for a specific scenario is a bit much, but fine.

src/CounterInterceptor/ClientCounter.cs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+public class ClientCounter
```
cretz, line 19: Why have this separate class? Why not just use `Dictionary<string, ClientCounts>` where it's needed? Do not use static state, store the dictionary on the interceptor and make it available to the creator of the interceptor.

src/CounterInterceptor/MyWorkflow.workflow.cs
```diff
@@ -0,0 +1,42 @@
...
+[Workflow]
+public class MyWorkflow
+{
+    private bool exit; // Automatically defaults to false
+
+    [WorkflowRun]
+    public async Task<string> RunAsync()
+    {
+        // Wait for greeting info
+        await Workflow.WaitConditionAsync(() => Name != null && Title != null);
```
cretz, line 14: These properties can never be null, so this statement has no value, it will always pass immediately

src/CounterInterceptor/MyActivities.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Diagnostics;
+using Temporalio.Activities;
+
+public class MyActivities
+{
+    [Activity]
+    public string SayHello(string name, string title)
+    {
+        return "Hello " + title + " " + name;
+    }
```
cretz, line 12:
```suggestion
    public string SayHello(string name, string title) => $"Hello {title} {name}";
```

Use string interpolation and single-statement syntax (will stop commenting on this, but it applies project wide)

src/CounterInterceptor/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+internal class Program
+{
+    private static async Task Main(string[] args)
+    {
+        var client = await TemporalClient.ConnectAsync(
+            options: new("localhost:7233")
+            {
+                Interceptors = new[]
+                {
+                    new SimpleClientCallsInterceptor(),
```
cretz, line 15: You should store this as a field and access the counters off of it instead of making counters static

src/CounterInterceptor/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+        var activities = new MyActivities();
+
+        var taskQueue = "CounterInterceptorTaskQueue";
+
+        var workerOptions = new TemporalWorkerOptions(taskQueue).
+                AddAllActivities(activities).
+                AddWorkflow<MyWorkflow>().
+                AddWorkflow<MyChildWorkflow>();
+
+        workerOptions.Interceptors = new[] { new SimpleCounterWorkerInterceptor() };
```
cretz, line 35: You should consider combining these interceptors. Make the one interceptor implement both `IClientInterceptor` _and_ `IWorkerInterceptor`. The it's automatically used for worker interceptors and you don't have to set this setting. This is what we recommend here and is documented on the `Interceptors` field for client. See the https://github.com/temporalio/samples-dotnet/tree/main/src/ContextPropagation sample.

src/CounterInterceptor/Program.cs
cretz:
```suggestion
            Console.WriteLine($"Workflow result is {result}");
```
or
```suggestion
            Console.WriteLine($"Workflow result is {0}", result);
```

But not both. This mistake is made in other lines below. Probably the former approach is best.

src/CounterInterceptor/README.md
```diff
@@ -0,0 +1,28 @@
...
+The sample demonstrates: 
+- the use of a Worker Workflow Interceptor that counts the number of Workflow Executions, Child Workflow Executions, and Activity Executions and the number of Signals and Queries. It is based
+off of the [Java sample](https://github.com/temporalio/samples-java/tree/main) located [here](https://github.com/temporalio/samples-java/tree/main/core/src/main/java/io/temporal/samples/countinterceptor)
+- the use of a Client Workflow Interceptor that counts the number of Workflow Executions and the number of Signals and Queries.
+
+## Start local Temporal Server
+```bash
+# run only once
+temporal server start-dev
+```
```
cretz, line 11:
```suggestion
To run, first see [README.md](https://github.com/temporalio/samples-dotnet/blob/main/README.md) for prerequisites
```

We like to keep the starting of the local server in one place since it can change and there are multiple options. This can be seen in other samples. (in fact, we will make these run easy on cloud too soon, so we don't want to have to come back and update these)

src/CounterInterceptor/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Run Worker Locally
```
cretz, line 13: This is more than the worker, it's the worker and the client calls mixed into one. Many samples split these, but don't have to here, but it's not just "worker" as the title suggests.

src/CounterInterceptor/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+            options: new("localhost:7233")
+            {
+                Interceptors = new[]
+                {
+                    new SimpleClientCallsInterceptor(),
+                },
+            });
+
+        using var tokenSource = new CancellationTokenSource();
+        Console.CancelKeyPress += (_, eventArgs) =>
```
cretz, line 20: This program is a run then complete, you don't need ctrl+c support as if it was a long running worker. I would recommend just removing cancel token and shutting down the worker manually if you must have the entire program self-contained (i.e. combining client and worker).

src/CounterInterceptor/README.md
```diff
@@ -0,0 +1,28 @@
...
+```bash
+# make sure you have temporal server running (see section above)
+dotnet run
+```
+
+## Run Worker using Temporal Cloud
```
cretz, line 19: I don't think this section applies anymore

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
```diff
@@ -0,0 +1,65 @@
+namespace TemporalioSamples.CounterInterceptor;
+
+using Temporalio.Client;
+using Temporalio.Client.Interceptors;
+
+public class SimpleClientCallsInterceptor : IClientInterceptor
+{
+    private ClientCounter clientCounter;
```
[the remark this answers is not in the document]
cretz: Bump

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        public override Task<WorkflowUpdateHandle<TResult>> StartWorkflowUpdateAsync<TResult>(
+            StartWorkflowUpdateInput input)
+        {
+            // not tracking this
+            return base.StartWorkflowUpdateAsync<TResult>(input);
+        }
+
+        private static string CheckId(string? id)
```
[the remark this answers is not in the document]
cretz: Bump, only three places have this, no need to make a whole new method for a simple `id ?? "None"` in those three places

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            var id = CheckId(input.Id);
+            root.clientCounter.AddQueryInvocation(id);
+            return base.QueryWorkflowAsync<TResult>(input);
+        }
+
+        public override Task<WorkflowUpdateHandle<TResult>> StartWorkflowUpdateAsync<TResult>(
+            StartWorkflowUpdateInput input)
+        {
+            // Not tracking this
```
cretz, line 50: Then can remove this method entirely

src/CounterInterceptor/SimpleCounterWorkerInterceptor.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        public override Task SignalChildWorkflowAsync(SignalChildWorkflowInput input)
+        {
+            return base.SignalChildWorkflowAsync(input);
+        }
+
+        public override Task SignalExternalWorkflowAsync(SignalExternalWorkflowInput input)
+        {
+            return base.SignalExternalWorkflowAsync(input);
+        }
```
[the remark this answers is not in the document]
cretz: Bump

src/CounterInterceptor/WorkerCounter.cs
cretz, line 1: All the same things on the `ClientCounter` class apply here

src/CounterInterceptor/ClientCounter.cs
rross: fixed.

src/CounterInterceptor/ClientCounter.cs
rross: Good catch. Fixed.

src/CounterInterceptor/ClientCounter.cs
rross: Done

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
rross: Removed

src/CounterInterceptor/ClientCounter.cs
rross: Refactored.

src/CounterInterceptor/WorkerCounter.cs
rross: Changed.

src/CounterInterceptor/MyActivities.cs
cretz: Bump, this was not addressed

src/CounterInterceptor/MyWorkflow.workflow.cs
cretz: Bump, this was never addressed

src/CounterInterceptor/Program.cs
cretz: Bump, this was never addressed

src/CounterInterceptor/Program.cs
cretz: Bump this was never addressed. If you are using interpolation, you don't _also_ pass as a param.

src/CounterInterceptor/README.md
cretz: Bump, this was never addressed

src/CounterInterceptor/README.md
cretz: Bump, this was never addressed

src/CounterInterceptor/README.md
cretz: Bump, this was never addressed

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
```diff
@@ -0,0 +1,97 @@
+namespace TemporalioSamples.CounterInterceptor;
+
+using Temporalio.Client;
+using Temporalio.Client.Interceptors;
+
+public record ClientCounts
```
cretz, line 6: You should only have top-level types named the same as the file in A non-Program.cs C# file.

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
```diff
@@ -0,0 +1,97 @@
...
+    public uint NumOfWorkflowExecutions(string workflowId) =>
+        clientDictionary[workflowId].Executions;
+
+    public uint NumOfSignals(string workflowId) =>
+        clientDictionary[workflowId].Signals;
+
+    public uint NumOfQueries(string workflowId) =>
+        clientDictionary[workflowId].Queries;
+
+    private void Add(string workflowId, string type)
```
cretz, line 44: This is not thread safe. What if multiple threads for the same workflow ID come in at the same time, there is a case where one dictionary value may get overwritten with another. Consider a `ConcurrentDictionary`.

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
```diff
@@ -0,0 +1,97 @@
...
+        private SimpleClientCallsInterceptor root;
+
+        public ClientOutbound(SimpleClientCallsInterceptor root, ClientOutboundInterceptor next)
+            : base(next) => this.root = root;
+
+        public override Task<WorkflowHandle<TWorkflow, TResult>> StartWorkflowAsync<TWorkflow, TResult>(
+            StartWorkflowInput input)
+        {
+            var id = input.Options.Id ?? "None";
+            root.Add(id, NumberOfWorkflowExecutions);
```
cretz, line 79: No need for these string constants, can change this to something like:

```suggestion
            Interlocked.Increment(ref root.ClientCounts(id).Executions);
```

Or

```suggestion
            root.UpdateClientCounts(id, counts => counts.Executions++);
```

Depending on how you want to do the thread safety

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
```diff
@@ -0,0 +1,97 @@
...
+        $"Total Number of Signals: {Signals}\n\t" +
+        $"Total Number of Queries: {Queries}";
+}
+
+public class SimpleClientCallsInterceptor : IClientInterceptor
+{
+    private const string NumberOfWorkflowExecutions = "numOfWorkflowExec";
+    private const string NumberOfSignals = "numOfSignals";
+    private const string NumberOfQueries = "numOfQueries";
+    private static Dictionary<string, ClientCounts> clientDictionary = new();
```
cretz, line 25: This should not be static and it should be readonly and you do not have to suffix "Dictionary", just call it `counts` or something

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
```diff
@@ -0,0 +1,97 @@
...
+
+    public ClientOutboundInterceptor InterceptClient(ClientOutboundInterceptor nextInterceptor) =>
+        new ClientOutbound(this, nextInterceptor);
+
+    public string Info() =>
```
cretz, line 30: May be easier if you just let the caller access the dictionary and do what they want with it instead of all of these separate methods to read/write it

src/CounterInterceptor/SimpleCounterWorkerInterceptor.cs
cretz, line 1: Most things in client interceptor class apply here

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
```diff
@@ -0,0 +1,97 @@
...
+
+    public uint Queries { get; internal set; }
+
+    public override string ToString() =>
+        $"\n\tTotal Number of Workflow Exec: {Executions}\n\t" +
+        $"Total Number of Signals: {Signals}\n\t" +
+        $"Total Number of Queries: {Queries}";
+}
+
+public class SimpleClientCallsInterceptor : IClientInterceptor
```
cretz, line 20: IMO for simplicity you should make one interceptor that implements both `IClientInterceptor` and `IWorkerInterceptor` (and you only then need to configure the client not the worker interceptor separately). You could also make just one counts object.

src/CounterInterceptor/SimpleCounterWorkerInterceptor.cs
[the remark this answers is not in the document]
cretz: Bump again, these weren't addressed

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
rross: Done

src/CounterInterceptor/SimpleCounterWorkerInterceptor.cs
rross: Done

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
rross: Done

src/CounterInterceptor/SimpleClientCallsInterceptor.cs
rross: Done.

src/CounterInterceptor/Counts.cs
```diff
@@ -0,0 +1,54 @@
+public record Counts
```
cretz, line 1: No benefit to making a record, might as well make a class. Also, make sure you put the namespace on this file.

src/CounterInterceptor/Counts.cs
```diff
@@ -0,0 +1,54 @@
...
+        clientExecutions = 0;
+        clientQueries = 0;
+        clientSignals = 0;
+        workflowExecutions = 0;
+        workflowSignals = 0;
+        workflowQueries = 0;
+        workflowChildExecutions = 0;
+        workflowActivityExecutions = 0;
+    }
+
```
cretz, line 14:
```suggestion
```

Not needed, this is the default value already of every field here.

src/CounterInterceptor/MyActivities.cs
```diff
@@ -0,0 +1,17 @@
+namespace TemporalioSamples.CounterInterceptor;
+
+using System.Diagnostics;
+using Temporalio.Activities;
+
+public class MyActivities
+{
+    [Activity]
+    public string SayHello(string name, string title) =>
+        "Hello " + title + " " + name;
```
cretz, line 10:
```suggestion
        $"Hello {title} {name}";
```

I think interpolation is a bit clearer here

src/CounterInterceptor/MyActivities.cs
```diff
@@ -0,0 +1,17 @@
...
+
+    [Activity]
+    public string SayGoodBye(string name, string title)
```
cretz, line 13: You are inconsistent with your single-line methods, this should be `=>` like the one just above it

src/CounterInterceptor/MyCounterInterceptor.cs
```diff
@@ -0,0 +1,137 @@
...
+using Temporalio.Client;
+using Temporalio.Client.Interceptors;
+using Temporalio.Worker.Interceptors;
+using Temporalio.Workflows;
+
+public class MyCounterInterceptor : IClientInterceptor, IWorkerInterceptor
+{
+    private ConcurrentDictionary<string, Counts> counts = new();
+
+    public ConcurrentDictionary<string, Counts> Counts => counts;
```
cretz, line 14:
```suggestion
    public ConcurrentDictionary<string, Counts> Counts { get; } = new();
```

Use auto properties

src/CounterInterceptor/MyCounterInterceptor.cs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private sealed class WorkflowInbound : WorkflowInboundInterceptor
+    {
+        private readonly MyCounterInterceptor root;
+
+        internal WorkflowInbound(MyCounterInterceptor root, WorkflowInboundInterceptor next)
+            : base(next) => this.root = root;
+
+        public override void Init(WorkflowOutboundInterceptor outbound)
```
cretz, line 76: Pedantic, but can also be single-method syntax using `=>`

src/CounterInterceptor/MyCounterInterceptor.cs
```diff
@@ -0,0 +1,137 @@
...
+            : base(next) => this.root = root;
+
+        public override Task<WorkflowHandle<TWorkflow, TResult>> StartWorkflowAsync<TWorkflow, TResult>(
+            StartWorkflowInput input)
+        {
+            var id = input.Options.Id ?? "None";
+            // Need to add an empty record if none exists
+            // we don't care if it doesn't add it as we will
+            // still increment the current value.
+            root.counts.TryAdd(id, new Counts());
```
cretz, line 49: This should happen on every call, not just start/execute workflow. The `root.counts[id].ClientSignals` call below will throw an exception when someone uses this to signal a workflow that wasn't started here. Same for other situations. Consider just having a helper on the root like:

```csharp
private void Increment(string id, Func<Counts, ref uint> toIncrement) =>
    Interlocked.Increment(ref toIncrement(counts.GetOrAdd(id, _ => new()));
```

Or something like that (untested). Then you can just change this to `root.Increment(id ?? "None", c => ref c.ClientExecutions);`.

src/CounterInterceptor/MyCounterInterceptor.cs
```diff
@@ -0,0 +1,137 @@
...
+            Interlocked.Increment(ref root.counts[id].ClientExecutions);
+            return base.StartWorkflowAsync<TWorkflow, TResult>(input);
+        }
+
+        public override Task SignalWorkflowAsync(SignalWorkflowInput input)
+        {
+            var id = input.Id ?? "None";
```
cretz, line 56: `input.Id` can never be null on this call (and some others), so `??` should be removed

src/CounterInterceptor/MyCounterInterceptor.cs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        public override Task<object?> ExecuteWorkflowAsync(ExecuteWorkflowInput input)
+        {
+            var id = Workflow.Info.WorkflowId;
+            // Need to add an empty record if none exists
+            // we don't care if it doesn't add it as we will
+            // still increment the current value.
+            root.counts.TryAdd(id, new Counts());
+            Interlocked.Increment(ref root.counts[id].WorkflowExecutions);
```
cretz, line 88: Note, this is not workflow executions, but workflow replays. You could get 1000 of these for the same workflow. Same for other counts in the workflow inbound/outbound. These will be double counted on every replay. You can consider not doing this on replay, or you can clarify in docs of the fields that they include replay attempts (you probably want the former).

src/CounterInterceptor/MyWorkflow.workflow.cs
cretz: Bump again, this is still not addressed

src/CounterInterceptor/Program.cs
cretz: Bump again, this is still not addressed

src/CounterInterceptor/README.md
cretz: Bump again, this is still not addressed

src/CounterInterceptor/README.md
cretz: Bump again, this is still not addressed

src/CounterInterceptor/MyCounterInterceptor.cs
rross: Oh wow. That's important to call out. 

I see that Workflow.Info has an Attempt field, but not sure it makes sense to use that to determine replay, as I think that is used for the number of retry attempts.

src/CounterInterceptor/MyWorkflow.workflow.cs
rross: Fixed.

src/CounterInterceptor/Program.cs
rross: Fixed.

src/CounterInterceptor/README.md
rross: Deleted.

src/CounterInterceptor/MyCounterInterceptor.cs
rross: Fixed

src/CounterInterceptor/MyCounterInterceptor.cs
rross: fixed

src/CounterInterceptor/MyCounterInterceptor.cs
cretz: There is `Workflow.Unsafe.IsReplaying` you can make sure is false before doing the increment (this is what we do in tracing interceptor for instance). It should be noted though that on task failure, this will run over and over in non-replaying situation, but besides task failure it will work.

src/CounterInterceptor/MyCounterInterceptor.cs
rross: I debated on whether it needed to be done on every call or not. 

I was able to use your example as you had it. Worked the first time. 

Thanks!

src/CounterInterceptor/MyCounterInterceptor.cs
rross: Fixed

src/CounterInterceptor/Counts.cs
rross: fixed

src/CounterInterceptor/Counts.cs
rross: Fixed

src/CounterInterceptor/MyActivities.cs
rross: Fixed

src/CounterInterceptor/MyActivities.cs
rross: Fixed

src/CounterInterceptor/Counts.cs
```diff
@@ -0,0 +1,43 @@
+namespace TemporalioSamples.CounterInterceptor;
+public class Counts
```
cretz, line 2: You are inconsistent with whether you add spacing between namespace and type declaration

cretz (commented):
Looks good, just want to take a look at the outstanding issues on the README

src/CounterInterceptor/README.md
rross: Finally addressed :)

src/CounterInterceptor/README.md
rross: Fixed.

src/CounterInterceptor/README.md
```diff
@@ -0,0 +1,13 @@
+# dotnet-counter-interceptor
```
cretz, line 1: Not really the best title, but that's ok

cretz (approved):
Looks good, thanks!

PR #87, opened 2024-10-07
Changed encrypted bytes to put tag last which matches the Java Sample

rross:
## What was changed
Changed the arrangement of segments in the encrypted bytes to match how Java does it. (Nonce, Encrypted Data, Tag)

## Why?
To provide encryption compatibility with our Java Encryption Sample, 

1. Closes 

Not applicable

2. How was this tested:

Tested in a different project to validate encryption is cross SDK compatible. Ran dotnet tests

3. Any docs updates needed?
No

### DIFF
--- a/src/Encryption/Codec/EncryptionCodec.cs
+++ b/src/Encryption/Codec/EncryptionCodec.cs
@@ -61,8 +61,8 @@ public sealed class EncryptionCodec : IPayloadCodec
 
     private byte[] Encrypt(byte[] data)
     {
-        // Our byte array will have a const-length nonce, const-length tag, and
-        // then the encrypted data. In real-world use, one may want to put nonce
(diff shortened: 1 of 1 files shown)


cretz:
Can you confirm what the behavior regarding tag placement is with https://github.com/temporalio/samples-typescript/tree/main/encryption, https://github.com/temporalio/samples-python/tree/main/encryption, and https://github.com/temporalio/samples-go/tree/main/encryption? At quick glance there is no tag on the latter two? Is it possible to confirm compatibility with this implementation and those 3 (ignoring Java for now)? I just want to confirm that .NET is the outlier and not Java.

cretz (approved):
Looks like this PR makes this sample work with multiple other-language samples repos. Marking approved. Merge at will.

PR #91, opened 2025-01-09
Update .NET SDK and add update-with-start samples

cretz:
## What was changed

* Update .NET SDK to 1.4.0
* Add sample for update-with-start early return
* Add sample for update-with-start lazy init

## Checklist

1. Closes #90

PR #92, opened 2025-01-10
Add F# examples

purkhusid:
## What was changed
Added a F# example that mirrors the ActivitySimple C# sample

## Why?
So that people that use F# can more easily see how Temporal code looks


cretz:
Thanks, this is great! I had some trouble with lambda expressions and F# in the past, happy to see this is working. I assume you manually confirmed it worked? Three requests:

* Update the primary `README.md` pointing to this directory as we do with the other samples
* Update `.editorconfig` to ignore CA1515 (or this PR can wait on https://github.com/temporalio/samples-dotnet/pull/91 to be merged where we did this at https://github.com/temporalio/samples-dotnet/pull/91/commits/23eecf7f4dbb494cb92c6f19c8202d7a7f996f16)
* Is it possible to write a test? Can follow what we do in `tests/ActivitySimple`? Ideally F# can be in the same project, but if not, maybe we need a new `tests-fsharp` project? If this is too much effort we can do in a separate issue.

purkhusid:
> Thanks, this is great! I had some trouble with lambda expressions and F# in the past, happy to see this is working. I assume you manually confirmed it worked? Three requests:
> 
>     * Update the primary `README.md` pointing to this directory as we do with the other samples
> 
>     * Update `.editorconfig` to ignore CA1515 (or this PR can wait on [Update .NET SDK and add update-with-start samples #91](https://github.com/temporalio/samples-dotnet/pull/91) to be merged where we did this at [23eecf7](https://github.com/temporalio/samples-dotnet/commit/23eecf7f4dbb494cb92c6f19c8202d7a7f996f16))
> 
>     * Is it possible to write a test? Can follow what we do in `tests/ActivitySimple`? Ideally F# can be in the same project, but if not, maybe we need a new `tests-fsharp` project? If this is too much effort we can do in a separate issue.

Thanks for taking a look. This works but I'm looking into adding a F# example for some of the more complicated examples as well. The overload resolution is quite painful in some cases so I might create a supporting library with F# utilities if I discover some nice patterns.

I'll also add some F# tests as well. I'll let you know once it's ready for review again.

cretz:
> The overload resolution is quite painful in some cases so I might create a supporting library with F# utilities if I discover some nice patterns.

I hit this too a while back (a couple of years ago in a [research branch](https://github.com/temporalio/samples-dotnet/compare/main...cretz:temporal-samples-dotnet:fsharp)). While I am not sure we're yet at a place where we want to maintain a `Temporalio.Extensions.FSharp` library in the primary SDK, putting generic extensions/utilities in a class in the sample here can help others. But don't feel obligated to do it here/now. And if we need to tweak the SDK in ways to make them work, we can discuss it.

PR #95, opened 2025-02-01
add sleep-for-days sample

THardy98:
added `sleep-for-days` sample

### DIFF
--- a/src/SleepForDays/Activities.cs
+++ b/src/SleepForDays/Activities.cs
@@ -1,0 +1,14 @@
+namespace TemporalioSamples.SleepForDays;
+
+using Microsoft.Extensions.Logging;
+using Temporalio.Activities;
+
+public class SleepForDaysActivities
+{
+    // Stub for an actual implementation for sending emails.
+    [Activity]
+    public void SendEmail(string msg)
+    {
+        ActivityExecutionContext.Current.Logger.LogInformation("{Msg}", msg);
+    }
+}
--- a/src/SleepForDays/Activities.cs
+++ b/src/SleepForDays/Activities.cs
@@ -3,7 +3,7 @@ namespace TemporalioSamples.SleepForDays;
 using Microsoft.Extensions.Logging;
 using Temporalio.Activities;
 
-public class SleepForDaysActivities
+public class Activities
 {
     // Stub for an actual implementation for sending emails.
     [Activity]
--- a/src/SleepForDays/SleepForDays.workflow.cs
+++ b/src/SleepForDays/SleepForDays.workflow.cs
@@ -1,0 +1,33 @@
+namespace TemporalioSamples.SleepForDays;
+
+using Microsoft.Extensions.Logging;
+using Temporalio.Workflows;
+
+[Workflow]
+public class SleepForDaysWorkflow
+{
+    private bool isComplete;
+
+    [WorkflowRun]
+    public async Task RunAsync()
+    {
+        isComplete = false;
+        while (!isComplete)
+        {
+            await Workflow.ExecuteActivityAsync(
+                (SleepForDaysActivities act) => act.SendEmail("Sleeping for 30 days"),
+                new() { StartToCloseTimeout = TimeSpan.FromSeconds(10) });
+            await Task.WhenAny(
+                Workflow.DelayAsync(TimeSpan.FromSeconds(30)),
+                Workflow.WaitConditionAsync(() => isComplete));
+        }
+
+        Workflow.Logger.LogInformation("done!");
+    }
+
+    [WorkflowSignal]
+    public async Task CompleteAsync()
+    {
+        isComplete = true;
+    }
+}
--- a/tests/SleepForDays/SleepForDaysWorkflowTests.cs
+++ b/tests/SleepForDays/SleepForDaysWorkflowTests.cs
@@ -1,0 +1,56 @@
+namespace TemporalioSamples.Tests.SleepForDays;
+
+using Temporalio.Activities;
+using Temporalio.Client;
+using Temporalio.Testing;
+using Temporalio.Worker;
+using TemporalioSamples.SleepForDays;
+using Xunit;
+using Xunit.Abstractions;
+
+public class SleepForDaysWorkflowTests : TestBase
+{
+    public SleepForDaysWorkflowTests(ITestOutputHelper output)
+        : base(output)
+    {
+    }
+
+    [Fact]
+    public async Task RunAsync_SleepForDays_Succeeds()
+    {
+        await using var env = await WorkflowEnvironment.StartTimeSkippingAsync();
+        var activities = new SleepForDaysActivities();
+
+        var activityExecutions = 0;
+        // Mock out the activity to assert number of executions
+        [Activity]
+        void SendEmail(string msg)
+        {
+            activityExecutions++;
+        }
+
+        using var worker = new TemporalWorker(
+            env.Client,
+            new TemporalWorkerOptions("sleep-for-days-task-queue").
+                AddActivity(SendEmail).
+                AddWorkflow<SleepForDaysWorkflow>());
+
+        var startTime = await env.GetCurrentTimeAsync();
+        await worker.ExecuteAsync(async () =>
+        {
+            var handle = await env.Client.StartWorkflowAsync(
+                (SleepForDaysWorkflow wf) => wf.RunAsync(),
+                new(id: $"wf-{Guid.NewGuid()}", taskQueue: worker.Options.TaskQueue!));
+            // Sleep for 90 days
+            await env.DelayAsync(TimeSpan.FromDays(90));
+            // Expect 3 activity executions
+            Assert.Equal(3, activityExecutions);
+            // Signal the workflow to complete
+            await handle.SignalAsync(wf => wf.CompleteAsync());
+            // Expect the same number of activity executions
+            Assert.Equal(3123123, activityExecutions);
+            // Assert at least 90 days of time have passed
+            Assert.True((await env.GetCurrentTimeAsync() - startTime) >= TimeSpan.FromDays(90));
+        });
+    }
+}
--- a/tests/SleepForDays/SleepForDaysWorkflowTests.cs
+++ b/tests/SleepForDays/SleepForDaysWorkflowTests.cs
@@ -14,7 +14,7 @@ public class SleepForDaysWorkflowTests : TestBase
     {
     }
 
-    [Fact]
+    [TimeSkippingServerFact]
     public async Task RunAsync_SleepForDays_Succeeds()
     {
         await using var env = await WorkflowEnvironment.StartTimeSkippingAsync();
--- a/tests/SleepForDays/SleepForDaysWorkflowTests.cs
+++ b/tests/SleepForDays/SleepForDaysWorkflowTests.cs
@@ -1,5 +1,6 @@
 namespace TemporalioSamples.Tests.SleepForDays;
 
+using Microsoft.Extensions.Logging;
 using Temporalio.Activities;
 using Temporalio.Client;
 using Temporalio.Testing;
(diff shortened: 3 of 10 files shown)


THardy98:
This is pretty much done - I'm just having trouble with the:
`await env.DelayAsync(TimeSpan.FromDays(90));`

it's not actually time-skipping.

I've made the assertion following that statement to fail, just to make sure that this doesn't pass CI.

src/SleepForDays/Activities.cs
```diff
@@ -0,0 +1,14 @@
+namespace TemporalioSamples.SleepForDays;
+
+using Microsoft.Extensions.Logging;
+using Temporalio.Activities;
+
+public class SleepForDaysActivities
```
cretz, line 6: Class name should match file name

src/SleepForDays/SleepForDays.workflow.cs
```diff
@@ -0,0 +1,33 @@
+namespace TemporalioSamples.SleepForDays;
+
+using Microsoft.Extensions.Logging;
+using Temporalio.Workflows;
+
+[Workflow]
+public class SleepForDaysWorkflow
```
cretz, line 7: Class name should match pre-extension part of file name (should probably change file name in this case)

src/SleepForDays/SleepForDays.workflow.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    private bool isComplete;
+
+    [WorkflowRun]
+    public async Task RunAsync()
+    {
+        isComplete = false;
```
cretz, line 14:
```suggestion
```

This is dangerous because it will overwrite a signal that comes before the workflow starts

src/SleepForDays/SleepForDays.workflow.cs
```diff
@@ -0,0 +1,33 @@
+namespace TemporalioSamples.SleepForDays;
+
+using Microsoft.Extensions.Logging;
+using Temporalio.Workflows;
+
+[Workflow]
+public class SleepForDaysWorkflow
+{
+    private bool isComplete;
```
cretz, line 9:
```suggestion
    private bool complete;
```

"is" prefix not usually needed in .NET

src/SleepForDays/SleepForDays.workflow.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    [WorkflowRun]
+    public async Task RunAsync()
+    {
+        isComplete = false;
+        while (!isComplete)
+        {
+            await Workflow.ExecuteActivityAsync(
+                (SleepForDaysActivities act) => act.SendEmail("Sleeping for 30 days"),
```
cretz, line 18: Can be a bit confusing to users for the send-email subject to be what the workflow is doing next, but not a big deal

src/SleepForDays/SleepForDays.workflow.cs
```diff
@@ -0,0 +1,33 @@
...
+                new() { StartToCloseTimeout = TimeSpan.FromSeconds(10) });
+            await Task.WhenAny(
+                Workflow.DelayAsync(TimeSpan.FromSeconds(30)),
+                Workflow.WaitConditionAsync(() => isComplete));
```
cretz, line 22: There is an overload of `WaitConditionAsync` that accepts a timeout

src/SleepForDays/SleepForDays.workflow.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        Workflow.Logger.LogInformation("done!");
+    }
+
+    [WorkflowSignal]
+    public async Task CompleteAsync()
+    {
+        isComplete = true;
+    }
```
cretz, line 32:
```suggestion
    public async Task CompleteAsync() => isComplete = true;
```

Can shorten some of these one-liners if you want

tests/SleepForDays/SleepForDaysWorkflowTests.cs
```diff
@@ -0,0 +1,56 @@
...
+                AddWorkflow<SleepForDaysWorkflow>());
+
+        var startTime = await env.GetCurrentTimeAsync();
+        await worker.ExecuteAsync(async () =>
+        {
+            var handle = await env.Client.StartWorkflowAsync(
+                (SleepForDaysWorkflow wf) => wf.RunAsync(),
+                new(id: $"wf-{Guid.NewGuid()}", taskQueue: worker.Options.TaskQueue!));
+            // Sleep for 90 days
+            await env.DelayAsync(TimeSpan.FromDays(90));
```
cretz, line 45: Are you sure you're not time skipping before the workflow is even waiting for the number of days? You may be time skipping before the workflow even reaches the wait condition.

src/SleepForDays/SleepForDays.workflow.cs
THardy98: We considered this, but figured having a direct `sleep` more clearly showed the ability to wait for long periods of time.

tests/SleepForDays/SleepForDaysWorkflowTests.cs
THardy98: I hadn't considered that. Is there a way to test for this? Maybe with a query?

src/SleepForDays/SleepForDays.workflow.cs
cretz: The concern here is that we usually encourage samples to look like a user would write them (otherwise we could have left timeout off of the wait condition calls). It also depends on if we're making a simple sample like for the website or not. If so and we want a clear "sleep" for this sample, we should consider using workflow cancellation as the interrupt mechanism because, even though we may not recommend cancellation as control flow, it makes the sample much more legible.

src/SleepForDays/SleepForDays.workflow.cs
```diff
@@ -0,0 +1,33 @@
...
+    [WorkflowRun]
+    public async Task RunAsync()
+    {
+        isComplete = false;
+        while (!isComplete)
+        {
+            await Workflow.ExecuteActivityAsync(
+                (SleepForDaysActivities act) => act.SendEmail("Sleeping for 30 days"),
+                new() { StartToCloseTimeout = TimeSpan.FromSeconds(10) });
+            await Task.WhenAny(
```
cretz, line 20:
```suggestion
            await Workflow.WhenAnyAsync(
```

Even though `Task.WhenAny(` technically works here in most cases, we encourage the safer wrapper

THardy98: Didn't realize there was a workflow wrapper, switched.

src/SleepForDays/SleepForDays.workflow.cs
THardy98: Shortened

src/SleepForDays/SleepForDays.workflow.cs
THardy98: Removed prefix.

src/SleepForDays/SleepForDays.workflow.cs
THardy98: Good point, didn't consider this. Don't think this is a _real_ concern with this sample, but I get that this might not be a pattern we want to demonstrate to new users. Removed.

src/SleepForDays/SleepForDays.workflow.cs
THardy98: Renamed file to `SleepForDaysWorkflow.workflow.cs`, looks a bit odd.

src/SleepForDays/Activities.cs
THardy98: Renamed class to `Activities`

src/SleepForDays/SleepForDays.workflow.cs
cretz: Yeah, though that same seemingly-odd name is present on similar files/workflows in this repo

src/SleepForDays/SleepForDays.workflow.cs
cretz: Make sure it's `WhenAnyAsync`

src/SleepForDays/SleepForDays.workflow.cs
THardy98: Yeah that's true. The intention is for this to be a reference for the web sample and the discussion on Slack leaned towards using the workflow function and stripping out the signal stuff if it's too complicated.

But imo I agree. I err on the side of no interrupt mechanism at all, would be the simplest to read for a new viewer. But we have some consensus on this already and some samples already merged. I don't think it's worth re-opening the discussion or being inconsistent with the other samples, especially given the scope.

cretz: Won't hold up this PR on this FWIW, but when we get to the public site content, I think it may not look like a sample

tests/SleepForDays/SleepForDaysWorkflowTests.cs
```diff
@@ -0,0 +1,56 @@
...
+    public SleepForDaysWorkflowTests(ITestOutputHelper output)
+        : base(output)
+    {
+    }
+
+    [Fact]
+    public async Task RunAsync_SleepForDays_Succeeds()
+    {
+        await using var env = await WorkflowEnvironment.StartTimeSkippingAsync();
+        var activities = new SleepForDaysActivities();
```
cretz, line 22: This is failing compilation

cretz (commented):
Also, make sure that this `.csproj` is added to the primary solution (`sln`) file. Can see what's there with `dotnet sln list` and can add with `dotnet sln add`. Also don't forget to add this sample to README list.

src/SleepForDays/SleepForDays.workflow.cs
THardy98: Yeah I think that's fair. Marketing team asked for PRs yesterday which I shared, with a little addendum that if they want simpler examples I can remove the signal handling stuff from the workflow and send them code snippets

tests/SleepForDays/SleepForDaysWorkflowTests.cs
THardy98: Forgot to push that change, whoops

src/SleepForDays/SleepForDays.workflow.cs
THardy98: Updated

THardy98:
Am I correct in thinking that `[TimeSkippingServerFact]` is used to avoid time skipping on platforms outside of x86/x64?
```
[xUnit.net 00:00:01.25]       Time-skipping test server only works on x86/x64 platforms
```
Saw this failure in CI when I had the `[Fact]` annotation and noticed the `[TimeSkippingServerFact]` annotation used in `ActivitySimple`'s workflow test.

tests/SleepForDays/SleepForDaysWorkflowTests.cs
```diff
@@ -41,14 +42,33 @@ await worker.ExecuteAsync(async () =>
...
+                    if (e.TimerStartedEventAttributes != null && (TimeSpan.FromDays(30) == e.TimerStartedEventAttributes.StartToFireTimeout.ToTimeSpan()))
+                    {
+                        timerStarted = true;
+                        break;
+                    }
+                }
+            }
+
+            // Sanity check - timer should have started.
+            Assert.True(timerStarted);
```
cretz, line 60: The final assert can never be false. Would recommend only looping a certain number of times and sleeping a short bit between each fetch. 

Actually, I just realized even in samples we have a `AssertMore.EventuallyAsync` helper, so you could probably just have:

```csharp
await AssertMore.EventuallyAsync(async () =>
{
    var history = await handle.FetchHistoryAsync();
    Assert.Contains(history.Events, e => e.TimerStartedEventAttributes != null);
});
```

(untested but the idea is there)

THardy98: Oh much nicer. I've also updated the activity execution count assertions. By the time the timer event for the sleep/delay is in history, we should have already executed the activity once. Skipping 90 more days should then give us 4 total executions not 3, which was erroneous. I wonder why that passed, if the timer existed in history before, then we still would have had to execute the activity, we still should have had 4 executions not 3.

cretz:
> Am I correct in thinking that [TimeSkippingServerFact] is used to avoid time skipping on platforms outside of x86/x64?
Yes

cretz (approved):
Looks great, feel free to merge if/when CI passes

THardy98:
TYFR!

PR #103, opened 2025-06-20
update infrequent polling example to use benign application exceptions

THardy98:
## What was changed
Use `BENIGN` application failures for infrequent polling activity

## Why?
Demonstrate the usage of `BENIGN` application failures for expected activity failures.

3. Any docs updates needed?
probably not

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -19,7 +19,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Temporalio" Version="1.6.0" />
+    <PackageReference Include="Temporalio" Version="1.7.0" />
     <PackageReference Include="Temporalio.Extensions.DiagnosticSource" Version="1.6.0" />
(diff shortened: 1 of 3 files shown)


src/Polling/Infrequent/TestService.cs
```diff
@@ -18,6 +20,9 @@ public async Task<string> GetServiceResultAsync(CancellationToken cancellationTo
             return "OK";
         }
 
-        throw new TestServiceException("Service is down");
+        // Set the error as BENIGN to indicate it is an expected error.
+        // BENIGN errors have activity failure logs downgraded to DEBUG level and
+        // do not emit activity failure metrics.
```
cretz, line 25: The test service is meant to be non-Temporal related. You wouldn't have control over the exception thrown by your upstream service. Rather, I think we should leave everything as is and just change `InfrequentPollingActivities` to throw a new benign exception wrapping this one.

THardy98: done

PR #109, opened 2025-09-16
Worker Deployment versioning sample

Sushisource:
## What was changed
Updating versioning samples in all langauges

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (commented):
Only minor things

src/WorkerVersioning/AutoUpgradingWorkflowV1.workflow.cs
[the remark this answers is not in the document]
Sushisource: Added a super obvious test that basically just runs the sample itself

tests/ConsoleWriter.cs
```diff
@@ -4,15 +4,28 @@ namespace TemporalioSamples.Tests;
...
-        output.WriteLine(value);
+        try
+        {
+            output.WriteLine(value);
+        }
+        catch (Exception ex) when (ex is InvalidOperationException && ex.Message.Contains("no currently active test"))
+        {
+            // Fall back to original console output when no test is active or writer is disposed
+            // This happens when background tasks write to console
+            originalOut.WriteLine(value);
```
Sushisource, line 28: Unclear why this is necessary. It's complaining on one of the console log lines in `RunDemoAsync`, which by definition finishes before the test does. I think there's some weird thread context switching going on or something. Open to better ways to fix this if you have any.

cretz: Hrmm, looking at https://github.com/xunit/xunit/issues/2377 and other links referencing that, can you try updating all 3 xunit deps in the tests csproj to latest (or at least 2.9.0) and see if this still occurs?

cretz (commented):
LGTM, just the one minor thing about seeing if newer xunit solves the console problem

cretz (approved):

Issue #101, opened 2025-06-17
[Feature Request] Environment configuration


THardy98:
**Describe the solution you'd like**

Update all samples to leverage external client configuration once https://github.com/temporalio/sdk-dotnet/issues/490 is complete.

PR #111, opened 2025-09-26
Standalone Client Environment Configuration

THardy98:
## What was changed
Added standalone client environment configuration sample.

## Why?
Display usage of client environment configuration sample.

1. Part of https://github.com/temporalio/samples-dotnet/issues/101

2. How was this tested:
Sample runs as intended

3. Any docs updates needed?
No


cretz (approved):
Usually would like at least one test, but since this is just making a simple client call, not really required

src/EnvConfig/LoadFromFile.cs
```diff
@@ -0,0 +1,54 @@
...
+            Console.WriteLine("\nAttempting to connect to client...");
+
+            var client = await TemporalClient.ConnectAsync(connectOptions);
+            Console.WriteLine("✅ Client connected successfully!");
+
+            // Test the connection by checking the service
+            var sysInfo = await client.Connection.WorkflowService.GetSystemInfoAsync(new());
+            Console.WriteLine("✅ Successfully verified connection to Temporal server!\n{0}", sysInfo);
+        }
+        catch (Exception ex) when (ex is not OperationCanceledException)
```
cretz, line 49:
```suggestion
        catch (Exception ex)
```

General question, why is operation canceled ok here? And really, why catch the exception at all?

PR #112, opened 2025-09-26
update samples to use env config

THardy98:
## What was changed
Updated existing samples to use client environment configuration.

## Why?
Client environment configuration is intended as the standard way to configure clients going forward.

1. Part of https://github.com/temporalio/samples-dotnet/issues/101

2. How was this tested:
Existing test suite passes

3. Any docs updates needed?
No

### DIFF
--- a/src/ActivityHeartbeatingCancellation/Program.cs
+++ b/src/ActivityHeartbeatingCancellation/Program.cs
@@ -1,17 +1,21 @@
 using Microsoft.Extensions.Logging;
 using Temporalio.Client;
+using Temporalio.Client.EnvConfig;
 using Temporalio.Exceptions;
 using Temporalio.Worker;
 using TemporalioSamples.ActivityHeartbeatingCancellation;
 
 // Create a client to localhost on default namespace
-var client = await TemporalClient.ConnectAsync(new("localhost:7233")
+var connectOptions = ClientEnvConfig.LoadClientConnectOptions();
+if (string.IsNullOrEmpty(connectOptions.TargetHost))
 {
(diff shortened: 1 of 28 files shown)


src/ActivityHeartbeatingCancellation/Program.cs
```diff
@@ -1,17 +1,21 @@
...
 using Temporalio.Client;
+using Temporalio.Client.EnvConfig;
 using Temporalio.Exceptions;
 using Temporalio.Worker;
 using TemporalioSamples.ActivityHeartbeatingCancellation;
 
 // Create a client to localhost on default namespace
-var client = await TemporalClient.ConnectAsync(new("localhost:7233")
+var connectOptions = ClientEnvConfig.LoadClientConnectOptions();
+if (string.IsNullOrEmpty(connectOptions.TargetHost))
```
cretz: Would recommend `connectOptions.TargetHost ??= "localhost:7233";` which I believe works. Also, I assume `connectOptions.TargetHost` is not set to an empty string if it is unset in env config, correct?

THardy98: > The class does accept null values, just not in that constructor, so you should use the other constructor, e.g. var options = new TemporalClientConnectOptions() { TargetHost = Address };

Ah I missed this.

> But I'd be ok if we do this for now and come back and ??= it when we represent Address as null properly. Would we prefer that approach? (if so, we can create an issue and I'll mark approved no this PR)

I'd prefer this. I can open a new PR on sdk and samples, would prefer to not be blocked on another release.

PR #133, opened 2026-04-12
Switch to using slnx instead of sln

### LINKED ISSUE #132: [Feature Request] Switch to using slnx instead of sln
### Is your feature request related to a problem? Please describe.

The .sln file format (TemporalSamples.sln) is legacy and a bit of a pain to deal with given this solution has so many projects and is always growing.

Replace with the new xml based [slnx](https://devblogs.microsoft.com/dotnet/introducing-slnx-support-dotnet-cli/) format. 

### Additional context

To upgrade run `dotnet sln migrate`

jakejscott:
## What was changed
- Upgraded to use dotnet 9.0
- Upgraded solution to use slnx file format
- Temporal nuget packages updated to 1.13.0
- Updated coverlet.collector to 8.0.1 because a warning was shown that it had a vulnerability. If this package is not actually used it probably could be removed.

## Checklist

1. Closes https://github.com/temporalio/samples-dotnet/issues/132

2. How was this tested:

- Ran `dotnet format--verify-no-changes`
- Ran `dotnet test`

3. Any docs updates needed?

- Readme was updated so that users know they need dotnet 9.


src/CounterInterceptor/MyCounterInterceptor.cs
```diff
@@ -136,9 +136,13 @@ internal ActivityInbound(MyCounterInterceptor root, ActivityInboundInterceptor n
 
         public override Task<object?> ExecuteActivityAsync(ExecuteActivityInput input)
         {
-            var info = ActivityExecutionContext.Current.Info;
-            var workflowId = info.IsWorkflowActivity ? info.WorkflowId! : throw new InvalidOperationException("Activity must be invoked from a workflow");
-            root.Increment(workflowId, c => Interlocked.Increment(ref root.Counts[workflowId].WorkflowActivityExecutions));
+            var id = ActivityExecutionContext.Current.Info.WorkflowId;
+
+            // TODO: What do we do here for standalone activities?
```
jakejscott: @cretz Left a note here because `WorkflowId` is now nullable due to standalone activities.

src/Mutex/Impl/MutexActivities.cs
```diff
@@ -25,14 +26,19 @@ public MutexActivities(ITemporalClient client)
     public async Task SignalWithStartMutexWorkflowAsync(SignalWithStartMutexWorkflowInput input)
     {
         var activityInfo = ActivityExecutionContext.Current.Info;
-        var workflowId = activityInfo.IsWorkflowActivity ? activityInfo.WorkflowId! : throw new InvalidOperationException("Activity must be invoked from a workflow");
 
-        await this.client.StartWorkflowAsync(
+        // TODO: What do we do here for standalone activities?
```
jakejscott: @cretz Also here

jmaeagle99:
Hi @jakejscott, thank you for your contribution. While I agree it would be great to update the solution file to use the slnx format, my understanding is that it is only support in the .NET 9+ SDKs. Since .NET 8 is still actively supported, I think it's best to keep the solution in a format that is compatible with all .NET 8+ SDKs for the time being.

PR #137, opened 2026-04-28
Nexus samples: use business IDs for workflow IDs

VegetarianOrc:
## What was changed

Stop using the Nexus request ID as the backing workflow ID. Build a meaningful business ID from the operation input.

## Why?

Our samples should reflect best practices around creating meaningful IDs. Part of https://github.com/temporalio/features/issues/692.


src/NexusCancellation/Handler/HelloService.cs
```diff
@@ -14,8 +14,11 @@ public class HelloService
...
-                    new() { Id = context.HandlerContext.RequestId }));
-}
\ No newline at end of file
+                    // dedupe workflow starts. For this example, use a business ID derived from
+                    // the greeting input so repeated operations for the same name and language
+                    // resolve to the same workflow.
+                    new() { Id = GetHelloWorkflowId(input) }));
+
+    private static string GetHelloWorkflowId(IHelloService.HelloInput input) =>
+        $"hello-{input.Language}-{input.Name.Trim().Replace(' ', '-')}";
```
chrsmith: The other examples, e.g. https://github.com/temporalio/samples-go/pull/460, aren't going through the `Name.Trim().Replace(...)` jazz.

Are spaces allowed in Workflow IDs? If they are, then maybe that would just be overkill for the sample, if that type of sanitization isn't required? (Or if it is, maybe it should be added to the other examples?)

jmaeagle99: I haven't found anything that states that the workflow ID cannot have spaces. If it operates with spaces, then let's remove the replacement so we don't lead customers into thinking its necessary.

jmaeagle99: Please remove the workflow ID sanitization from all of the samples that were added in this PR.

VegetarianOrc: ~Sorry for the delay here! I've removed the unnecessary sanitization here.~

Will update remaining.

jmaeagle99 (approved):

PR #152, opened 2026-05-29
Lambda Worker Sample

eamsden:
Needs temporalio/sdk-dotnet#721

## What was changed
Add sample code for a worker running on AWS Lambda

## Why?
SDK-level implementation of serverless workers on AWS Lambda

## Checklist

1. Closes 

2. How was this tested:

Deployed a Lambda worker and ran a workflow

3. Any docs updates needed?


src/LambdaWorker/Function.cs
```diff
@@ -0,0 +1,25 @@
...
+using Temporalio.Extensions.Aws.Lambda.OpenTelemetry;
+
+public class LambdaFunction
+{
+    private static readonly Func<object?, ILambdaContext, Task> WorkerHandler =
+        TemporalLambdaWorker.CreateHandler(
+            new WorkerDeploymentVersion(
+                LambdaWorkerSample.DeploymentName,
+                LambdaWorkerSample.BuildId),
+            Configure);
```
02strich: nit: is this how you expect people to style (with the separate `Configure`)?

src/LambdaWorker/Starter/Program.cs
02strich, line 1: is this the starter? is it normal in .NET to have this just as another cs file in the same folder?

eamsden: Yes, this is the Workflow starter. Addressed by moving it into a separate Starter project and making the Lambda worker a library, so starter-only code is not included in the Lambda artifact.

src/LambdaWorker/Function.cs
eamsden: Addressed: the Lambda handler now uses an inline configuration callback, matching the canonical SDK examples.

src/LambdaWorker/README.md
```diff
@@ -0,0 +1,299 @@
+# Lambda Worker
```
jmaeagle99, line 1: No best place to put this comment... I think the structure of this sample can be improved. As it is, looking at the source, it just looks like a pile of commingled files. Take a look at https://github.com/temporalio/samples-dotnet/tree/main/src/Encryption as an example of clean separation. Maybe the following structure:
- `Deploy` (contains all of the scripts)
- `Starter` (same as the current Starter directory)
- `Worker` (all of the *.csproj and *.cs files for the worker, rename project file to have `.Worker` on end, remember to update deploy-lambda.sh with that change too, the compile exclusion in the csproj is removable)
-  `README.md` remains at the root
- `*.sample` files remain at the root, but maybe rename to `*.template.yaml` if they are to be copy/pasted with substitutions.

TemporalioSamples.sln
```diff
@@ -97,6 +97,10 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TemporalioSamples.NexusMult
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TemporalioSamples.EnvConfig", "src\EnvConfig\TemporalioSamples.EnvConfig.csproj", "{52CE80AF-09C3-4209-8A21-6CFFAA3B2B01}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TemporalioSamples.LambdaWorker", "src\LambdaWorker\TemporalioSamples.LambdaWorker.csproj", "{D1D6B7AD-0E12-4D8C-83BE-9CE3EB916DD7}"
```
jmaeagle99: These two projects should be grouped into a folder in this solution file, probably called `LambdaWorker`.

tests/LambdaWorker/LambdaWorkerTests.cs
```diff
@@ -0,0 +1,85 @@
...
+        Assert.Contains(
+            options.Activities,
+            activity => activity.MethodInfo?.DeclaringType == typeof(Activities));
+        var workflow = options.Workflows.Single(
+            workflow => workflow.Type == typeof(SampleWorkflow));
+        Assert.Equal(VersioningBehavior.Pinned, workflow.VersioningBehavior);
+    }
+
+    [Fact]
+    public void ConfigureWorkerOptions_UsesEnvironmentOverrides()
```
jmaeagle99: I don't think this test is actually testing anything about the sample itself. Feels like it's testing the product or even just the `EnvironmentVariableScope` behavior, which isn't used anywhere else other than this test) rather than demonstrating any correctness about the sample.

src/LambdaWorker/Deploy/enable-telemetry.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
jmaeagle99, line 1: Should this file have a `.sh` extension? Maybe also rename to `enable-telemetry.sh`, or something more indicative of what this file does?

src/LambdaWorker/README.md
```diff
@@ -0,0 +1,299 @@
...
+This command supplies all environment variables together because the Lambda
+`--environment` option replaces the complete variable map. Its output is
+restricted so the API key is never echoed:
+
+```bash
+aws lambda update-function-configuration \
+  --function-name "$FUNCTION_NAME" \
+  --timeout 60 \
+  --layers "$ADOT_LAYER_ARN" \
+  --environment "Variables={TEMPORAL_API_KEY=${TEMPORAL_API_KEY},SSL_CERT_FILE=/etc/pki/tls/certs/ca-bundle.crt,OPENTELEMETRY_COLLECTOR_CONFIG_URI=/var/task/otel-collector-config.yaml,TEMPORAL_TASK_QUEUE=${TASK_QUEUE},TEMPORAL_LAMBDA_WORKFLOW_ID_PREFIX=${WORKFLOW_ID_PREFIX},TEMPORAL_LAMBDA_DEPLOYMENT_NAME=${DEPLOYMENT_NAME},TEMPORAL_LAMBDA_BUILD_ID=${BUILD_ID}}" \
```
jmaeagle99: I don't think that TEMPORAL_LAMBDA_WORKFLOW_ID_PREFIX is needed here since only the starter project consumes it.

src/LambdaWorker/README.md
```diff
@@ -0,0 +1,299 @@
...
+
+You can also use Workers > Deployments > Actions > Validate Connection in the
+Temporal Cloud UI.
+
+## 7. Start a Workflow
+
+Run the separate starter project:
+
+```bash
+TEMPORAL_CONFIG_FILE="$PWD/temporal.toml" \
```
jmaeagle99, line 220: Need to set TEMPORAL_LAMBDA_WORKFLOW_ID_PREFIX here according to how the starter code functions. Or maybe not since there is a fallback.

src/LambdaWorker/README.md
```diff
@@ -0,0 +1,299 @@
...
+  mise exec dotnet@8 -- dotnet run \
```
jmaeagle99: Why invoke via mise here and not for the `dotnet publish` call in the deploy-lambda.sh script? I think remove mise since it's not a prereq and doesn't add value here.

src/LambdaWorker/Function.cs
```diff
@@ -0,0 +1,25 @@
+// @@@SNIPSTART dotnet-lambda-worker
```
jmaeagle99: Why snipsync here? I thought we were applying docs snipsync in temporalio/features? Which allows us to evolve the samples without breaking docs.

src/NexusContextPropagation/NexusContextPropagationInterceptor.cs
```diff
@@ -61,4 +61,4 @@ public override Task<NexusWorkflowOperationHandle<TResult>> ScheduleNexusOperati
             return base.ScheduleNexusOperationAsync<TResult>(input);
         }
     }
-}
\ No newline at end of file
+}
```
jmaeagle99: nit: undo unrelated change

tests/LambdaWorker/LambdaWorkerTests.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            var result = await env.Client.ExecuteWorkflowAsync(
+                (SampleWorkflow wf) => wf.RunAsync("Serverless Lambda Worker!"),
+                new(id: $"workflow-{Guid.NewGuid()}", taskQueue: worker.Options.TaskQueue!));
+            Assert.Equal("Hello, Serverless Lambda Worker!!", result);
+        });
+    }
+
+    [Fact]
+    public void ConfigureWorkerOptions_SetsExpectedWorkerRegistration()
```
jmaeagle99: This also feels like testing SDK behavior rather than sample behavior. I would drop it.

src/LambdaWorker/README.md
```diff
@@ -0,0 +1,299 @@
...
+- .NET 8, `zip`, and `openssl`
+- An ADOT Collector Lambda layer ARN for the target region and architecture;
+  see the
+  [ADOT Lambda documentation](https://aws-otel.github.io/docs/getting-started/lambda#getting-started-with-aws-lambda-layers)
+
+Run the commands below from `src/LambdaWorker`.
+
+## Files
+
+| File | Description |
```
jmaeagle99: This can be drastically simplified with the directory refactor mentioned above. I don't know if you really need to call out each individual file either. Maybe with the refactor, just state what each directory contains in general.

jmaeagle99 (approved):

PR #175, opened 2026-07-13
Batch example

### LINKED ISSUE #131: [Sample Request] Batch

Like Java batch at https://github.com/temporalio/samples-java/tree/main/core/src/main/java/io/temporal/samples/batch but in dotnet

ddieruf:
## What was changed

Added a `.NET` port of the Java `batch` samples (`samples-java/core/src/main/java/io/temporal/samples/batch`), under a new `src/Batch` directory with one project per pattern:

- **Iterator** — processes a page of records at a time via child Workflows run in parallel, continuing-as-new between pages.
- **HeartbeatingActivity** — processes the whole batch inside a single heartbeating Activity, resuming from the last recorded offset after a worker restart instead of starting over.
- **SlidingWindow** — keeps a fixed number of record-processing child Workflows running in parallel at all times, using completion Signals and continue-as-new to keep history bounded indefinitely; multiple partitions run side by side via `BatchWorkflow` for higher total throughput.

Each sample follows this repo's existing conventions (`dotnet run worker` / `dotnet run workflow`, per-sample README, `TemporalioSamples.sln` entries) and mirrors the Java sample's file/class structure closely so it's easy to compare side by side.

One deliberate deviation from the Java version: `SlidingWindowBatchWorkflow` continues-as-new on `Workflow.ContinueAsNewSuggested` in addition to a fixed page size, rather than relying on the fixed count alone. A fixed count can be outgrown by other accumulated history (e.g. completion Signals) before it's reached — this pattern is based on how we handle the same problem in production. A `TestContinueAsNew` input flag makes that branch deterministically testable without generating a large amount of real history.

## Why?

There was no .NET equivalent of these samples, which came up in a community discussion about batch-processing workarounds (paging through large datasets, bounding history size, safely reporting child-Workflow completion across continue-as-new). This fills that gap using patterns we've validated in a production import pipeline.

## Checklist

1. Closes #131 

2. How was this tested:
   - `dotnet build` / `dotnet format --verify-no-changes` on the full solution.
   - Added automated tests under `tests/Batch/*` (xUnit, using `WorkflowEnvironment.StartTimeSkippingAsync` for the Iterator and SlidingWindow Workflow tests, and `ActivityEnvironment` for a targeted heartbeat-resume test on HeartbeatingActivity). All pass.
   - Manually ran each sample's worker + starter against a local Temporal dev server end to end and inspected results/history via the `temporal` CLI. This caught and fixed a real bug during development: `SlidingWindowBatchWorkflow`'s original return value summed to double the actual record count across partitions; it now returns the Signal-tracked `progress` count instead.

3. Any docs updates needed?
   - Added `src/Batch/README.md` (overview) plus a README per sample.
   - Added a `Batch` entry to the root `README.md` sample index.
   - No docs.temporal.io updates needed.


src/Batch/Iterator/RecordLoaderActivities.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// Returns the next page of records.
+    /// </summary>
+    /// <param name="pageSize">Maximum number of records to return.</param>
+    /// <param name="offset">Offset of the next page.</param>
+    /// <returns>Empty list if there are no more records to process.</returns>
+    [Activity]
+    public static IReadOnlyList<SingleRecord> GetRecords(int pageSize, int offset)
+    {
+        var records = new List<SingleRecord>(pageSize);
+        if (offset < pageSize * PageCount)
```
jmaeagle99, line 28: This sample set size being a function of the pageSize is slightly odd. I would suggest changing the PageCount const to a TotalCount and just compare offset to TotalCount, just like you did in SlidingWindow.

src/Batch/SlidingWindow/BatchWorkflow.workflow.cs
```diff
@@ -0,0 +1,60 @@
...
+    [WorkflowRun]
+    public async Task<int> RunAsync(int pageSize, int slidingWindowSize, int partitions)
+    {
+        // The sample partitions the dataset into contiguous ranges. A real application can
+        // choose any other way to divide the records into multiple collections.
+        var totalCount = await Workflow.ExecuteActivityAsync(
+            () => RecordLoaderActivities.GetRecordCount(),
+            new() { StartToCloseTimeout = TimeSpan.FromSeconds(5) });
+
+        var partitionSize = (totalCount / partitions) + (totalCount % partitions > 0 ? 1 : 0);
```
jmaeagle99, line 32: nit: this looks like ceiling division, which can lead to the last partition being empty depending on the inputs. So it might start a child workflow that does nothing. Not necessarily a bug but an unnecessary execution.

For example, totalCount = 5 and partitions = 4 gives a partitionSize of 2. Those partitions are [0, 2), [2, 4], [4, 5), and [6, 5). The last one is clearly empty.

src/Batch/SlidingWindow/SlidingWindowBatchWorkflow.workflow.cs
```diff
@@ -0,0 +1,219 @@
...
+[Workflow]
+public class SlidingWindowBatchWorkflow
+{
+    /// <summary>
+    /// Accumulates records to remove for signals delivered before <see cref="RunAsync"/> starts
+    /// executing. This is the workaround for the race between continue-as-new and a child
+    /// reporting completion: a signal can be delivered to the new run before its workflow run
+    /// method has had a chance to populate <see cref="currentRecords"/> from the input.
+    /// </summary>
+    private readonly HashSet<int> recordsToRemove = new();
```
jmaeagle99, line 24: Take a look at WorkflowInit, which would allow defining a constructor with the same arguments as the WorkflowRun method. Might be able to set `currentRecords` from the constructor and have the signal handler and run methods use that field without having to store extra state.

src/Batch/SlidingWindow/SlidingWindowBatchWorkflow.workflow.cs
```diff
@@ -0,0 +1,219 @@
...
+        trackedRecords.ExceptWith(recordsToRemove);
+        progress += countBefore - trackedRecords.Count;
+
+        var pageSize = input.PageSize;
+        var offset = input.Offset;
+        var slidingWindowSize = input.SlidingWindowSize;
+
+        // For ease of testing, forces continue-as-new well before the real history-size
+        // threshold would suggest it, instead of relying on Workflow.ContinueAsNewSuggested
+        // alone. Mirrors ClusterManagerWorkflow's maxHistoryLength in the SafeMessageHandlers
```
jmaeagle99, line 66: Please remove the reference to the SafeMessageHandlers sample. It doesn't matter what the other samples do, and it's quite possible they will naturally diverge over time. Leaving this type of comment might cause confusion if that divergence occurs.

src/Batch/README.md
```diff
@@ -0,0 +1,16 @@
...
+   workflows, continuing-as-new between pages.
+2. [Heartbeating Activity](./HeartbeatingActivity/README.md) - Processes the whole batch inside a
+   single heartbeating Activity.
+3. [Sliding Window](./SlidingWindow/README.md) - Keeps a fixed number of record processing child
+   workflows running in parallel at all times, using Signals and continue-as-new to keep the
+   Workflow history size bounded indefinitely.
+
+These are .NET ports of the
+[Java batch samples](https://github.com/temporalio/samples-java/tree/main/core/src/main/java/io/temporal/samples/batch),
+kept as close as possible to the original Java structure and naming.
```
jmaeagle99, line 16: Please remove "kept as close as possible to the original Java structure and naming". That shouldn't bear weight on the .NET samples. The samples can diverge over time and this line is nearly promising that they will not.

src/Batch/SlidingWindow/README.md
```diff
@@ -0,0 +1,42 @@
...
+single record. When a child completes, a new child is immediately started.
+
+A parent Workflow calls continue-as-new after starting a preconfigured number of children. A
+child completion is reported through a Signal, as a parent cannot directly wait for a child that
+was started by a previous run.
+
+Multiple instances of `SlidingWindowBatchWorkflow` run in parallel, each processing a subset of
+records, to support a higher total rate of processing.
+
+This is the sample that demonstrates the workaround for the race between continue-as-new and a
```
jmaeagle99, line 16: This set of comments will need to be updated if the `WorkflowInit` feature is employed.

jmaeagle99 (commented):
Thanks for contributing these samples. Wanted to provide some initial feedback while I go through the rest of it.

jmaeagle99 (commented):
Also, in the PR description, you mention "which came up in a community discussion about batch-processing workarounds". Do you have a link to the community discussion that you could share? Or you could drop it into the T[emporal Community dotnet-sdk Slack channel](https://temporalio.slack.com/archives/C012SHMPDDZ).

ddieruf:
@jmaeagle99 thank you for the feedback (sorry for the late reply). We have this work in the queue. Should have the PR revised soon.

PR #200, opened 2026-07-27
Improve Guidance on when to use Sync/Async Nexus Operations

JoshuaFrenchwood:
## What was changed

Improved the NexusSimple sample to highlight when to use Synchronous vs Asynchronous Nexus Operations. Feedback on the nexus dotnet sdk told us that this was not clear enough. So we are improving this in our examples and documentation.  The sample now clearly describes when to use sync or async.
## Why?

This was requested feedback from users of the dotnet sdk
## Checklist

1. Closes 

2. How was this tested:
Tested locally, samples still work when running locally now.
3. Any docs updates needed?


src/NexusSimple/README.md
```diff
@@ -1,8 +1,22 @@
...
+contrasts synchronous and asynchronous Nexus operations:
 
-### Instructions
+* `Echo` uses `OperationHandler.Sync` because it is a short, bounded, in-memory operation with no
+  external calls or side effects.
+* `SayHello` is backed by a Workflow that waits on a durable 15-second timer before returning. The
+  delay intentionally exceeds the 10-second synchronous Nexus request deadline.
+
+Use `OperationHandler.Sync` only for highly reliable, low-latency, bounded operations that complete
+well within that short request window. Prefer an asynchronous operation when latency or availability is uncertain,
```
Quinn-With-Two-Ns:
```suggestion
well within that short request window. Use an asynchronous operation when latency or availability is uncertain, 
```

I'd use a bit stronger langauge since it is really a requirment

src/NexusSimple/README.md
```diff
@@ -1,8 +1,22 @@
...
+* `Echo` uses `OperationHandler.Sync` because it is a short, bounded, in-memory operation with no
+  external calls or side effects.
+* `SayHello` is backed by a Workflow that waits on a durable 15-second timer before returning. The
+  delay intentionally exceeds the 10-second synchronous Nexus request deadline.
+
+Use `OperationHandler.Sync` only for highly reliable, low-latency, bounded operations that complete
+well within that short request window. Prefer an asynchronous operation when latency or availability is uncertain, 
+the work might exceed the handler deadline, or execution depends on a potentially unreliable service or database.
+
+The asynchronous handler must still start the Workflow and return an operation token within
```
Quinn-With-Two-Ns: > still start the Workflow

Could we be bit more generic here? Since we will have more async operations very soon like stand alone activities or Workflow Update and just use workflow as an example?

JoshuaFrenchwood: Agreed, updated this to be more generic

Quinn-With-Two-Ns (approved):

PR #214, opened 2026-08-21
Add Nexus SAA sample

atol:
## What was changed
Adds a sample demonstrating using `TemporalOperationHandler` to start a standalone Activity.

### DIFF
--- a/src/NexusStandaloneActivity/Handler/GreetingService.cs
+++ b/src/NexusStandaloneActivity/Handler/GreetingService.cs
@@ -3,22 +3,25 @@ namespace TemporalioSamples.NexusStandaloneActivity.Handler;
 using NexusRpc.Handlers;
 using Temporalio.Nexus;
 
+// Declares a single operation whose backing execution is a Standalone Activity.
 [NexusServiceHandler(typeof(IGreetingService))]
 public class GreetingService
 {
-    [NexusOperationHandler]
-    public IOperationHandler<IGreetingService.GreetingInput, IGreetingService.GreetingOutput> Greet() =>
-        // TemporalOperationHandler maps a Temporal execution onto a Nexus Operation. Here the execution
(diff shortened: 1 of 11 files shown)


src/NexusStandaloneActivity/Handler/GreetingService.cs
```diff
@@ -0,0 +1,24 @@
+namespace TemporalioSamples.NexusStandaloneActivity.Handler;
+
+using NexusRpc.Handlers;
+using Temporalio.Nexus;
+
+[NexusServiceHandler(typeof(IGreetingService))]
+public class GreetingService
+{
+    [NexusOperationHandler]
```
Quinn-With-Two-Ns: We can simplify this a lot by using `TemporalOperation`. Removes the whole `TemporalOperationHandler.FromHandleFactory`

src/NexusStandaloneActivity/IGreetingService.cs
```diff
@@ -0,0 +1,16 @@
+namespace TemporalioSamples.NexusStandaloneActivity;
+
+using NexusRpc;
+
+// Nexus service definition shared by the caller and the handler. It declares a single operation
```
jmaeagle99: nit: I don't think the interface should document what backs the operation implementation and should be moved to said operation implementation, namely the `GreetingService`.

jmaeagle99 (commented):
Other than the following comment and the feedback in https://github.com/temporalio/samples-dotnet/pull/214/changes#r3846380545, I think this sample looks good.

src/NexusStandaloneActivity/Handler/GreetingService.cs
atol: Thanks! Updated to use `TemporalOperation`.

jmaeagle99 (approved):

PR #220, opened 2026-08-28
Enable transitive pinning

jmaeagle99:
## What was changed

Enable central package transitive pinning so that transitive package versions are controlled by the `Directory.Packages.props` file.

## Why?

Centrally manage transitive dependency versions and avoid PRs like #216

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
-    <CentralPackageTransitivePinning>true</CentralPackageTransitivePinning>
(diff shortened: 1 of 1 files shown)


Directory.Packages.props
```diff
@@ -1,6 +1,7 @@
 <Project>
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <CentralPackageTransitivePinning>true</CentralPackageTransitivePinning>
```
eamsden: Looks like this isn't the correct property?

https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#transitive-pinning

eamsden (approved):

PR #225, opened 2026-09-08
Update Nexus messaging samples to use Temporal Operation handler

atol:
## What was changed
* Update Nexus messaging samples to use Temporal Operation handler
* Add example for Signal-with-Start

## How was this tested
Ran the samples and verified output

### DIFF
--- a/tests/NexusMessaging/CallerPatternTests.cs
+++ b/tests/NexusMessaging/CallerPatternTests.cs
@@ -1,6 +1,7 @@
 namespace TemporalioSamples.Tests.NexusMessaging;
 
 using Temporalio.Client;
+using Temporalio.Testing;
 using Temporalio.Worker;
 using TemporalioSamples.NexusMessaging.CallerPattern.Caller;
 using TemporalioSamples.NexusMessaging.CallerPattern.Handler;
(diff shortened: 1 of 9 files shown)


tests/NexusMessaging/CallerPatternTests.cs
```diff
@@ -1,30 +1,48 @@
...
     [Fact]
     public async Task RunAsync_CallerWorkflow_Succeeds()
     {
+        // SetLanguage is backed by a Workflow Update, which needs a dev server build that supports
+        // update callbacks.
+        await using var env = await WorkflowEnvironment.StartLocalAsync(new()
+        {
+            DevServerOptions = new()
+            {
+                DownloadVersion = "v1.7.4-standalone-nexus-operations",
```
Evanthx, line 28: I am wondering if we should break this version out into a constant somewhere - it is in this two times, and is in the repo a few more times. But that might be better as a separate change, especially as we should do that across all the sample repos.

Evanthx (approved):
Looks OK, one comment but not something that needs to change in this PR!

tests/NexusMessaging/CallerPatternTests.cs
jmaeagle99: I'd be fine leaving as is for now, especially if it matches the other values. Let's refactor it later.

